PR #6, opened 2018-05-20
basic data_device helpers

### LINKED ISSUE #3: Design an abstraction for DataDevices (copy/paste and drag'n'drop)
The root global of the data device tree is the [`wl_data_device_manager`](https://docs.rs/wayland-client/0.20.1/wayland_client/protocol/wl_data_device_manager/index.html).

I'm not yet very familiar with this part of the protocol.

vberger:
Fixes #3

### DIFF
--- a/examples/selection.rs
+++ b/examples/selection.rs
@@ -1,14 +1,17 @@
 extern crate byteorder;
 extern crate smithay_client_toolkit as sctk;
 
-use std::cmp::min;
-use std::io::{BufWriter, Seek, SeekFrom, Write};
+use std::io::{BufWriter, Read, Seek, SeekFrom, Write};
 use std::sync::{Arc, Mutex};
 
 use byteorder::{NativeEndian, WriteBytesExt};
 
 use sctk::Environment;
+use sctk::data_device::{DataDevice, DndEvent, ReadPipe};
 use sctk::keyboard::{map_keyboard_auto, Event as KbEvent};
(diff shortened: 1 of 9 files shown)


examples/selection.rs
```diff
@@ -0,0 +1,192 @@
...
+                        }
+                    });
+                }
+            }
+            _ => (),
+        }
+    });
+
+    if !env.shell.needs_configure() {
+        // initial draw to bottstrap on wl_shell
```
Drakulix: *bottstrap

src/env.rs
```diff
@@ -141,6 +144,11 @@ impl Environment {
                 shm_formats2.lock().unwrap().push(format);
             });
 
+        let data_device_manager = manager
+            .instantiate_auto::<wl_data_device_manager::WlDataDeviceManager>()
+            .expect("Server didn't advertize `wl_data_device_manager`?!")
```
Drakulix: So this will make winit apps panic on smithay right now, doesn't it? :D 
Why don't you treat at least some of the objects of env optional?

vberger: Well, yes, if I release a new version of SCTK right away, it will. ^^

The main reason I didn't make them optional is that they are part of the core protocol: any compositor is supposed to support them. As such, it allows me to remove the burden of handling an `Option<_>` from the API.

Though, it may be possible to add some sneaky logic to lazily instantiate these globals upon use, I need to think about it. But this would be the scope for an other PR I think.

PR #7, opened 2018-06-13
Fix build on OpenBSD

farhaven:
Nix 0.10.0 does not work on OpenBSD, but 0.11.0 does.

Signed-off-by: Gregor Best <gbe@unobtanium.de>


farhaven:
Wayland doesn't exist on OpenBSD, but this crate is a dependency of the glutin crate.

vberger:
Thanks!

vberger:
Version 0.2.3 is published with the dependency update.

PR #9, opened 2018-07-12
Remove borders from windows

trimental:
I think that following the lead of GTK, QT..etc borders around windows should not be rendered as it uses up screen space and can look ugly. 

This pull request changes the drawing of the decorations so that the sides and bottom borders become invisible (it doesn't remove them because they are needed for window resizing). 
screenshot from 2018-07-12 14-09-11

DECORATION_SIZE now controls how far away your mouse can be from the window to resize it (warning that it still plays a part in the positioning of buttons though)

Because of the adjustments to window geometry which is calculated with headless_geometry() the compositor can snap the inner surface to the sides of the screen and show it correctly in the gnome activities view
screenshot from 2018-07-12 14-10-26

Full screen works without borders as you would expect any other window to do
screenshot from 2018-07-12 14-10-39

I've also tested that it works with decorations set to hidden.

I can only test on gnome so I'm unsure whether these changes will work properly for all compositors and I didn't really spend a lot of time testing windows with other settings but it seems to work perfectly with alacritty and the selection example. I set this as default behavior that cannot be changed as I couldn't think of any downsides, although it would be easy to make this a setting with something like set_borders(bool). Thanks

### DIFF
--- a/src/window/basic_frame.rs
+++ b/src/window/basic_frame.rs
@@ -393,19 +393,19 @@ impl Frame for BasicFrame {
                 // For every pixel in top border
                 for y in 0..DECORATION_TOP_SIZE {
                     for _ in 0..DECORATION_SIZE {
-                        writer.write_u32::<NativeEndian>(0x00_00_00_00);
+                        let _ = writer.write_u32::<NativeEndian>(0x00_00_00_00);
                     }
                     for _ in 0..width {
-                            writer.write_u32::<NativeEndian>(color);
+                        let _ = writer.write_u32::<NativeEndian>(color);
                     }
                     for _ in 0..DECORATION_SIZE {
-                        writer.write_u32::<NativeEndian>(0x00_00_00_00);
+                        let _ = writer.write_u32::<NativeEndian>(0x00_00_00_00);
                     }
                 }
 
                 // For every pixel in the other borders
                 for _ in DECORATION_TOP_SIZE * (width + 2 * DECORATION_SIZE)..pxcount {
-                    writer.write_u32::<NativeEndian>(0x00_00_00_00);
+                    let _ = writer.write_u32::<NativeEndian>(0x00_00_00_00);
                 }
 
                 draw_buttons(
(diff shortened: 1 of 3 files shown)


src/window/basic_frame.rs
```diff
@@ -543,26 +558,24 @@ impl Frame for BasicFrame {
...
-                width + 2 * (DECORATION_SIZE as i32),
-                height + DECORATION_SIZE as i32 + DECORATION_TOP_SIZE as i32,
-            )
+            (width, height + DECORATION_TOP_SIZE as i32)
         }
     }
 
-    fn location(&self) -> (i32, i32) {
-        (-(DECORATION_SIZE as i32), -(DECORATION_TOP_SIZE as i32))
+    fn headless_geometry(&self, x: i32, y: i32, width: i32, height: i32) -> (i32, i32, i32, i32) {
```
vberger: This trait is a public interface, so adding a new required method to it is a breaking change.

I'm not sure I see why this method is needed though, couldn't you have reached the same result by changing the implementation of `location()` to this?

```rust
fn location(&self) -> (i32, i32) {
    if self.hidden { (0, 0) } else { (0, -DECORATION_TOP_SIZE as i32) }
}
```

src/window/basic_frame.rs
```diff
@@ -390,9 +390,24 @@ impl Frame for BasicFrame {
             // draw the grey background
             {
                 let mut writer = BufWriter::new(&mut *pool);
-                for _ in 0..pxcount {
-                    let _ = writer.write_u32::<NativeEndian>(color);
+                // For every pixel in top border
+                for y in 0..DECORATION_TOP_SIZE {
+                    for _ in 0..DECORATION_SIZE {
+                        writer.write_u32::<NativeEndian>(0x00_00_00_00);
```
vberger: These writes generate compilation warnings, could you silence them with `let _ = ...` please?

trimental: both have been fixed in the cleanup commit

vberger:
Ah, I forgot, could you just add an entry in the changelog, in the `unreleased` section?

Other than that, all is good.

trimental:
updated the change log

vberger:
Perfect, thanks!

PR #10, opened 2018-07-12
Fix the display of buttons on small window sizes

trimental:
Forgot to update the sizes that buttons are no longer displayable


vberger:
Ah yes, right. But the offset should be `2*ds`, as borders were removed from both sides of the window.

PR #12, opened 2018-07-12
Seperate header bar and top border

trimental:
### Changes
* Added a header bar part to basic frame and made top a border piece above the header bar for resizing
* Change the way the buttons are laid out so that changes to the HEADER_SIZE (DECORATION_TOP_SIZE) or BORDER_SIZE (DECORATION_SIZE) don't effect their layout
* Rename DECORATION_SIZE and DECORATION_TOP_SIZE as they are very similar, long and slightly unclear
* Border size is set to 12 to match the similar distance that gtk has
* I ran a cargo fmt just on basic_frame.rs

### Effects
* The header bar size and border size can now be any value without unintended effect to the window (they can be set instead of const if wanted)
* The header bar can be better and properly sized from the top at all header bar sizes

I wasn't sure whether this was changelog worthy so I didn't add it but these changes do slightly change the way the window is resized from the top so I can put in the changelog if you want.

### DIFF
--- a/src/window/basic_frame.rs
+++ b/src/window/basic_frame.rs
@@ -26,9 +26,9 @@ use utils::{DoubleMemPool, MemPool};
  * Drawing theme definitions
  */
 
-const DECORATION_SIZE: u32 = 12;
-const DECORATION_TOP_SIZE: u32 = 32;
-const BUTTON_BORDER_SPACE: u32 = 10;
+const BORDER_SIZE: u32 = 12;
+const HEADER_SIZE: u32 = 32;
+const BUTTON_SPACE: u32 = 10;
 
 // defining the color scheme
 const INACTIVE_BORDER: u32 = 0xFF606060;
@@ -144,9 +144,9 @@ fn precise_location(old: Location, width: u32, x: f64, y: f64) -> Location {
         Location::Head | Location::Button(_) => find_button(x, y, width),
 
         Location::Top | Location::TopLeft | Location::TopRight => {
-            if x <= DECORATION_SIZE as f64 {
+            if x <= BORDER_SIZE as f64 {
                 Location::TopLeft
-            } else if x >= (width + DECORATION_SIZE) as f64 {
+            } else if x >= (width + BORDER_SIZE) as f64 {
                 Location::TopRight
             } else {
                 Location::Top
@@ -154,9 +154,9 @@ fn precise_location(old: Location, width: u32, x: f64, y: f64) -> Location {
         }
 
         Location::Bottom | Location::BottomLeft | Location::BottomRight => {
-            if x <= DECORATION_SIZE as f64 {
+            if x <= BORDER_SIZE as f64 {
                 Location::BottomLeft
-            } else if x >= (width + DECORATION_SIZE) as f64 {
+            } else if x >= (width + BORDER_SIZE) as f64 {
                 Location::BottomRight
             } else {
                 Location::Bottom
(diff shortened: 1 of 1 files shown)


trimental:
When reviewing it might be helpful to change this
```
// For every pixel in borders
for _ in HEADER_SIZE * (width + 2 * BORDER_SIZE)..pxcount {
    let _ = writer.write_u32::<NativeEndian>(0x00_00_00_00);
```
to this
```
// For every pixel in borders
for _ in HEADER_SIZE * (width + 2 * BORDER_SIZE)..pxcount {
    let _ = writer.write_u32::<NativeEndian>(0xee_ee_00_00);
```
so you can see the borders of the window

trimental:
I forgot to add that my reasoning for separating the header and top bars is because it is the standard behaviour of qt and gtk as it allows the whole headbar to be dragged and clicked while allowing the window to be resized from the top

vberger:
I tried to pull you branch an run the `image_viewer` example, and go this failure:

```
wl_shm_pool@10: error 1: invalid width, height or stride (12x332, 48)
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value:
Os { code: 71, kind: Other, message: "Protocol error" }'
```

It looks like your computations of buffer sizes or stride is wrong somewhere.

trimental:
screenshot from 2018-07-12 22-00-09

strange works for me and a git pull says I'm upto date, does this happen just on image viewer? can you try selection

trimental:
I'm sure that I've got a up to date and working copy of my fork, its either that you are running a modified version of image viewer, a wrong clone of my fork or that maybe mutter is more accepting of buffer sizes and stride errors than the compositor you are using?

vberger:
It depends on which image I load, some work and some don't. There might be a subtle error in the dimensions calculation that does not always make the result wrong?

One image that fails for me is 300x300, an other is 500x524, however I have a 432x277 that works.

trimental:
I'll get a 300x300 image to investigate

trimental:
It seems to work with 300x300 images now after a change in the calculation of total pixels, it panicked because the right buffer would try to access bytes in the pool not allocated by resize(pxcount)

src/window/basic_frame.rs
```diff
@@ -390,21 +402,21 @@ impl Frame for BasicFrame {
...
             {
                 let mut writer = BufWriter::new(&mut *pool);
-                // For every pixel in top border
-                for y in 0..DECORATION_TOP_SIZE {
-                    for _ in 0..DECORATION_SIZE {
+                // For every pixel in header
+                for y in 0..HEADER_SIZE {
+                    for _ in 0..BORDER_SIZE {
                         let _ = writer.write_u32::<NativeEndian>(0x00_00_00_00);
                     }
```
vberger: I think the lateral transparent drawings on the header are redundant with the lateral subsurfaces, no?

trimental: Unless I understood you wrong or I didn't test properly, the transparent color of the borders won't overwrite the color of the headers when layered on top of each other.

trimental: Actually I can implement the positioning and change the width of the header so the drawing of transparent color is redundant

vberger: I just meant that the header subsurface does not need to be `HEADER_SIZE + 2*BORDER_SIZE` large, but only `HEADER_SIZE`. Isn't that the case?

trimental: Yeah and also the positioning needs to change from
```
 (-(BORDER_SIZE as i32), -(HEADER_SIZE as i32)
```
to
```
(0, -(HEADER_SIZE as i32)
```
otherwise it renders color to the left of the window.

trimental: Does the resize and position headerbar commit do what you were thinking of

src/window/basic_frame.rs
```diff
@@ -429,123 +433,160 @@ impl Frame for BasicFrame {
...
+            self.inner.parts[HEAD]
                 .subsurface
-                .set_position(-(DECORATION_SIZE as i32), -(DECORATION_TOP_SIZE as i32));
+                .set_position(0, -(HEADER_SIZE as i32));
+            self.inner.parts[HEAD].surface.attach(Some(&buffer), 0, 0);
+            if self.surface_version >= 4 {
+                self.inner.parts[HEAD].surface.damage_buffer(
+                    0,
+                    0,
+                    (width + 2 * BORDER_SIZE) as i32,
```
vberger: we only need to damage `width` here

src/window/basic_frame.rs
```diff
@@ -429,123 +433,160 @@ impl Frame for BasicFrame {
...
+                    HEADER_SIZE as i32,
+                );
+            } else {
+                // surface is old and does not support damage_buffer, so we damage
+                // in surface coordinates and hope it is not rescaled
+                self.inner.parts[HEAD].surface.damage(
+                    0,
+                    0,
+                    (width + 2 * BORDER_SIZE) as i32,
```
vberger: same here, only need to damage `width`

trimental:
Would it be best to handle the write_u32s with unwraps or expects so that if writing does fail you can see where it failed, or maybe 
```
if write_u32::<NativeEndian>(0x00_00_00_00).is_err() {
    return
}
```
to stop the redrawing of surfaces if the program has failed to write to the pool.

src/window/basic_frame.rs
trimental: I also updated the damage to the side buffers as I changed their height to extend for the full height of the window

vberger:
Thanks, I'll merge this! :+1: 

Regarding the result, I guess the reasonable thing to do would be to change the `Frame::redraw()` method to return an `io::Result<()>`, and the `Window::refresh()` method to forward it directly, and let the user decide what to do in case the drawing of the decorations failed...

PR #13, opened 2018-07-13
Round window corners

trimental:
### Changes
* Window corners are rounded by calculating circle_width (X) from y and also ROUNDING_SIZE which sets the radius.
- diagram
drawing

### Effects
* Windows look more standard as gtk, qt and x11 also round their windows by default
* The const ROUNDING_SIZE can be used to set the amount of rounding, I found that without shading 3 works best however once shading is done it will probably increase to 4-8 like gtk.

### Issues
* Corners can look slightly pixelated as I have yet to try and implement window shading.
* At a ROUNDING_SIZE of higher than 3 the window doesn't fit nicely in the corners of mutter (should be fixed with shading).

### Improvements
* The calculation of circle_width could be cached instead of recalculated on every redraw however all redraw calculations aren't cached I think.

I think both the rounded and 'boxy' style look alright (it could maybe even be a setting as ROUNDING_SIZE=0 sets the boxy style), however as rounded is more standard with gtk, qt, x11, I would say it is a better default.

- Comparing gtk, client-toolkit and client toolkit rounded
screenshot from 2018-07-13 17-52-24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - BasicFrame: remove side and bottom border decorations
+- BasicFrame: round window corners
 
 ## 0.2.5 -- 2018-07-10
 
--- a/src/window/basic_frame.rs
+++ b/src/window/basic_frame.rs
@@ -29,6 +29,7 @@ use utils::{DoubleMemPool, MemPool};
 const BORDER_SIZE: u32 = 12;
 const HEADER_SIZE: u32 = 32;
 const BUTTON_SPACE: u32 = 10;
+const ROUNDING_SIZE: u32 = 3;
 
 // defining the color scheme
 const INACTIVE_BORDER: u32 = 0xFF606060;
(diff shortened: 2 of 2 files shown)


vberger:
Wow, this looks pretty neat!

The rounding radius is small so the effect is subtle, but it really does make a difference.

Thanks a lot for the time you are investing into making these border prettier!

vberger:
Btw, do you have more pending changes you are working on regarding the borders? I think your 3 PRs are worth a point release, but I'd feel silly to release it now just to learn you have a fourth PR incoming. :sweat_smile:

trimental:
I don’t have any current work on the borders. I was next going to work on buttons and colouring schemes to make the windows look hopefully a bit more natural and native, then try to add shading around the windows like gtk does (or at least around the corners). I would say now might be a good point release as I don’t know how long the three things I mentioned will take. ps I’m glad you like my changes as this and wayland rs are the first projects I’ve ever contributed to 😀

PR #14, opened 2018-07-14
Window drawing performance

trimental:
Given that thousands of pixels are written on resize I found that changing the constants manually instead of using byteorder to change them for every pixel has a fair impact on performance when redrawing the window. These changes on my computer leads to a 30ms (1.1x) increase in performance on debug and 4-6ms (3-4x) increase in performance on release. Its worth noting that because the drawing of the window is much much faster on release you can barely notice the difference of the 3x speed increase

### DIFF
--- a/src/window/basic_frame.rs
+++ b/src/window/basic_frame.rs
@@ -31,7 +31,6 @@ const BUTTON_SPACE: u32 = 10;
 const ROUNDING_SIZE: u32 = 3;
 
 // defining the color scheme
-
 #[cfg(target_endian = "little")]
 mod colors {
     pub const INACTIVE_BORDER: &[u8] = &[0x60, 0x60, 0x60, 0xFF];
@@ -44,7 +43,6 @@ mod colors {
     pub const YELLOW_BUTTON_HOVER: &[u8] = &[0x40, 0xFF, 0xFF, 0xFF];
     pub const YELLOW_BUTTON_DISABLED: &[u8] = &[0x20, 0x80, 0x80, 0xFF];
 }
-
 #[cfg(target_endian = "big")]
 mod colors {
     pub const INACTIVE_BORDER: &[u8] = &[0xFF, 0x60, 0x60, 0x60];


trimental:
Also I have a question that relates to performance, can users access the pool that is used to for drawing the window decorations?

trimental:
I have added the sharing of pixel data between the borders for speed as it now only draws the bear minimum of pixels needed, although I may have to revert this for window shading. Both changes give me 11ms (12x) speed increase per redraw on release.

src/window/basic_frame.rs
```diff
@@ -32,15 +30,15 @@ const BUTTON_SPACE: u32 = 10;
...
-const YELLOW_BUTTON_DISABLED: u32 = 0xFF808020;
+const INACTIVE_BORDER: &[u8] = &[0x60, 0x60, 0x60, 0xFF];
+const ACTIVE_BORDER: &[u8] = &[0x80, 0x80, 0x80, 0xFF];
+const RED_BUTTON_REGULAR: &[u8] = &[0x40, 0x40, 0xB0, 0xFF];
+const RED_BUTTON_HOVER: &[u8] = &[0x40, 0x40, 0xFF, 0xFF];
+const GREEN_BUTTON_REGULAR: &[u8] = &[0x40, 0xB0, 0x40, 0xFF];
+const GREEN_BUTTON_HOVER: &[u8] = &[0x40, 0xFF, 0x40, 0xFF];
+const YELLOW_BUTTON_REGULAR: &[u8] = &[0x40, 0xB0, 0xB0, 0xFF];
+const YELLOW_BUTTON_HOVER: &[u8] = &[0x40, 0xFF, 0xFF, 0xFF];
+const YELLOW_BUTTON_DISABLED: &[u8] = &[0x20, 0x80, 0x80, 0xFF];
```
vberger: Okay, I didn't expect `byteorder` to be such a bottleneck.

However, hard-coding the values like this is kind of wrong, given the wayland protocol specifies that values should be in native endianness. So this would be wrong if run in a big endian architecture (not sure if that could happen though ?). In all cases, to be safe, I'd rather keep handling endianness correctly. Things like `#[cfg(target_endian = "little")]` should allow this I believe.

src/window/basic_frame.rs
```diff
@@ -387,9 +385,12 @@ impl Frame for BasicFrame {
...
             // resize the pool as appropriate
-            let pxcount = 4 * (2 * (BORDER_SIZE * (width + 2 * BORDER_SIZE))
-                + 2 * (BORDER_SIZE * (height + 2 * BORDER_SIZE))
-                + (HEADER_SIZE * width));
+            let pxcount = if (width + 2 * BORDER_SIZE) > (height + HEADER_SIZE) {
+                (HEADER_SIZE * width) + ((width + 2 * BORDER_SIZE) * BORDER_SIZE)
+            } else {
+                (HEADER_SIZE * width) + ((height + HEADER_SIZE) * BORDER_SIZE)
+            };
+
```
vberger: This can be factored using `std::cmp::max()`, and would improve readability I believe:

```rust
use std::cmp::max;

let pxcount = HEADER_SIZE * width +
              BORDER_SIZE * max(height + HEADER_SIZE, width + 2 * BORDER_SIZE);
```

src/window/basic_frame.rs
```diff
@@ -1,8 +1,6 @@
 use std::io::{BufWriter, Seek, SeekFrom, Write};
 use std::sync::{Arc, Mutex};
 
-use byteorder::{NativeEndian, WriteBytesExt};
```
vberger: Is the `byteorder` crate used elsewhere? If not it should be removed from the dependencies.

vberger: if they are only used in the examples, it can be downgraded to a dev-dependency then.

src/window/basic_frame.rs
trimental: I chose to keep the constants in little endian and reverse if big because the device will more then likely be little endian so its faster

vberger: I was rather thinking about defining the constants using `#[cfg(...)]` to avoid runtime costs in all cases, so something like:

```rust
#[cfg(target_endian = "little")]
mod colors {
    /* color definitions in little endian */
}
#[cfg(target_endian = "big")]
mod colors {
    /* color definitions in big endian */
}
use self::colors::*;
```

PR #15, opened 2018-07-14
Add a example to test performance

trimental:
I think it would be nice to have a way to quickly test the performance of the client-toolkit, the performance example allows testing the time it takes for the window frame to redraw as well the window contents (the frame is the main thing but the contents are measured to show perspective). You press the space bar to get the performance measurements


trimental:
After thinking about it I am deciding to close since this example doesn't really display anything to users, I think I'll just use performance.rs on my local copy instead

PR #16, opened 2018-07-15
Add key repetition

trimental:
This pull request adds key repetition as a possible utility for the keyboard module. The work isn't there yet for release as there will at least need to be a setting for enabling/disabling it and bitflags to control what keys are repeatable. I'm pull requesting to see your views on the key repetition and whether it would be suitable for the client toolkit. I saw the pull request for winit but that looked pretty stale and I think it would fit better as an option in the client toolkit then in winit. I used threads, channels and sleeping as opposed to something like tokio as its simpler and lighter, the only downside I can think of is that threads might last a few hundred milliseconds longer due to sleeping (which I'm not sure whether that matters?) however cpu wise it should be pretty light. I don't know how releases work but maybe it would be a minor breakage if it was defaulted to off?

- [x] Interact with secondary thread using chan crate
- [x] Add parameter to key mapping to control key repetition
- [x] Add the ability to identify repeated key events
- [x] Allow the control of what types of keys are repeated
- [x] Fix shift + characters
- [x] Fix use of is_ascii_punctuation
- [x] Update examples
- [x] Make repeated key events member objects dynamic instead of static (eg time)
- [x] Fix updating of rand dependency (from chan) to a version not supported on rust 1.20
- [x] Dynamic state when repeating keys

### DIFF
--- a/src/keyboard/mod.rs
+++ b/src/keyboard/mod.rs
@@ -691,8 +691,8 @@ where
                                     },
                                     proxy.clone()
                                 );
-                                let delay = chan::after_ms(thread_repeat_timing.1);
-                                let tick = chan::tick_ms(thread_repeat_timing.0);
+                                let delay = chan::after_ms(repeat_timing.1);
+                                let tick = chan::tick_ms(repeat_timing.0);
                                 chan_select! {
                                     delay.recv() => {},
                                     thread_kill_r.recv() => { return }
--- a/src/keyboard/mod.rs
+++ b/src/keyboard/mod.rs
@@ -605,25 +605,12 @@ where
 
                         // Start a thread that spawns the same key event in intervals
                         thread::spawn(move || {
-                            thread_user_impl.lock().unwrap().receive(
-                                Event::Key {
-                                    serial,
-                                    time,
-                                    modifiers,
-                                    rawkey: key,
-                                    keysym: sym,
-                                    state: key_state,
-                                    utf8: utf8.clone(),
-                                },
-                                proxy.clone()
-                            );
+                            let mut first = true;
                             // Check if the key is repeatable
                             if let Some(utf8) = utf8 {
                                 let utf8_key = utf8.as_bytes()[0] as char;
                                 if utf8_key.is_ascii_graphic() || utf8_key == ' ' || utf8_key == 8 as char {
                                     let repeat_timing = thread_repeat_timing.lock().unwrap();
-                                    // Pause before repeation
-                                    thread::sleep(Duration::from_millis(repeat_timing.1));
                                     loop {
                                         // Break if been sent a kill request
                                         if let Ok(_) = reciever.try_recv() {
@@ -646,7 +633,12 @@ where
                                             break;
                                         }
                                         // Interval length
-                                        thread::sleep(Duration::from_millis(repeat_timing.0));
+                                        if first {
+                                            thread::sleep(Duration::from_millis(repeat_timing.1));
+                                            first = false;
+                                        } else {
+                                            thread::sleep(Duration::from_millis(repeat_timing.0));
+                                        }
                                     }
                                 }
                             };
--- a/src/keyboard/mod.rs
+++ b/src/keyboard/mod.rs
@@ -22,6 +22,7 @@ use std::ptr;
 use std::sync::{mpsc, Arc, Mutex};
 use std::thread;
 use std::time::Duration;
+use std::time::Instant;
 
 use memmap::MmapOptions;
 
@@ -712,6 +713,7 @@ where
                                     },
                                     proxy.clone(),
                                 );
+                                let time_tracker = Instant::now();
                                 // Delay
                                 thread::sleep(Duration::from_millis(repeat_timing.1));
                                 match thread_kill_chan.lock().unwrap().1.try_recv() {
@@ -719,10 +721,11 @@ where
                                     _ => {}
                                 }
                                 loop {
+                                    let elapsed_time = time_tracker.elapsed();
                                     thread_user_impl.lock().unwrap().receive(
                                         Event::Key {
                                             serial,
-                                            time,
+                                            time: time + elapsed_time.as_secs() as u32 * 1000 + elapsed_time.subsec_millis(),
                                             modifiers,
                                             rawkey: key,
                                             keysym: sym,
@@ -757,7 +760,7 @@ where
                         }
                     } else if key_held == Some(key) {
                         // If key released then send a kill message to the thread
-                        kill_chan.lock().unwrap().0.send(());
+                        kill_chan.lock().unwrap().0.send(()).unwrap();
                         kill_chan = Arc::new(Mutex::new(mpsc::channel::<()>()));
                         key_held = None;
                         user_impl.lock().unwrap().receive(
--- a/src/keyboard/mod.rs
+++ b/src/keyboard/mod.rs
@@ -858,9 +858,16 @@ where
                     let modifiers = state.mods_state.clone();
 
                     if key_state == wl_keyboard::KeyState::Pressed {
+                        let repeatable = unsafe {
+                            if (XKBH.xkb_keymap_key_repeats)(state.xkb_keymap, sym) == 1 {
+                                true
+                            } else {
+                                false
+                            }
+                        };
                         match key_repeat_kind {
                             KeyRepeatKind::Fixed {..} | KeyRepeatKind::System {..} => {
-                                if let Some(utf8) = utf8.clone() {
+                                if repeatable {
                                     key_held = Some(key);
                                     // Clone variables for the thread
                                     let thread_kill_chan = kill_chan.clone();
@@ -880,7 +887,7 @@ where
                                             rawkey: key,
                                             keysym: sym,
                                             state: key_state,
-                                            utf8: Some(utf8.clone()),
+                                            utf8: utf8.clone(),
                                         },
                                         proxy,
                                     );
@@ -903,7 +910,7 @@ where
                                                     rawkey: key,
                                                     keysym: sym,
                                                     state: key_state,
-                                                    utf8: Some(utf8.clone()),
+                                                    utf8: utf8.clone(),
                                                 }
                                             );
                                             // Rate
--- a/src/keyboard/mod.rs
+++ b/src/keyboard/mod.rs
@@ -353,26 +353,6 @@ impl Drop for KbState {
     }
 }
 
-bitflags! {
-    /// A bitflag of the types of keys to be repeated
-    pub struct RepeatedKeyTypes: u32 {
-        /// alphabetic characters a-z A-Z
-        const ALPHABETIC = 0b1;
-        /// numeric characters 0-9
-        const NUMERIC = 0b10;
-        /// alphabetic and numeric characters
-        const ALPHANUMERIC = Self::ALPHABETIC.bits | Self::NUMERIC.bits;
-        /// space character ' '
-        const SPACE = 0b100;
-        /// backspace character
-        const BACKSPACE = 0b1000;
-        /// punctuation characters !-~
-        const PUNCTUATION = 0b10000;
-        /// alphabetic, numeric, space, backspace and punctuation characters
-        const ALL = Self::ALPHABETIC.bits | Self::NUMERIC.bits | Self::SPACE.bits | Self::BACKSPACE.bits | Self::PUNCTUATION.bits;
-    }
-}
-
 /// Determines the behaviour of key repetition
 #[derive(PartialEq)]
 pub enum KeyRepeatKind {
@@ -384,14 +364,9 @@ pub enum KeyRepeatKind {
         rate: u64,
         /// delay (in milisecond) between a key press and the start of repetition
         delay: u64,
-        /// a bitflag of the types of keys to be repeated
-        key_types: RepeatedKeyTypes,
     },
     /// keys will be repeated at a rate and delay set by the wayland server
-    System {
-        /// a bitflag of the types of keys to be repeated
-        key_types: RepeatedKeyTypes,
-    },
+    System,
 }
 
 #[derive(Debug)]
@@ -463,8 +438,6 @@ pub enum Event<'a> {
         ///
         /// will always be `None` on key release events
         utf8: Option<String>,
-        /// identifies the event as original or repeated
-        repeated: bool,
     },
     /// Repetition information advertizing
     RepeatInfo {
@@ -473,6 +446,26 @@ pub enum Event<'a> {
         /// delay (in milisecond) between a key press and the start of repetition
         delay: i32,
     },
+}
+
+/// An event sent at repeated intervals for keys determined by xkb_keymap_key_repeats
+pub struct KeyRepeatEvent {
+    /// serial number of the event
+    pub serial: u32,
+    /// time at which the keypress occured
+    pub time: u32,
+    /// current state of the modifiers
+    pub modifiers: ModifiersState,
+    /// raw value of the key
+    pub rawkey: u32,
+    /// interpreted symbol of the key
+    pub keysym: u32,
+    /// new state of the key
+    pub state: KeyState,
+    /// utf8 interpretation of the entered text
+    ///
+    /// will always be `None` on key release events
+    pub utf8: Option<String>,
 }
 
 /// Implement a keyboard to automatically detect the keymap
@@ -487,7 +480,6 @@ pub enum Event<'a> {
 /// Returns an error if xkbcommon could not be initialized.
 pub fn map_keyboard_auto<Impl>(
     keyboard: NewProxy<wl_keyboard::WlKeyboard>,
-    key_repeat_kind: KeyRepeatKind,
     implementation: Impl,
 ) -> Result<Proxy<wl_keyboard::WlKeyboard>, (Error, NewProxy<wl_keyboard::WlKeyboard>)>
 where
@@ -500,8 +492,7 @@ where
     Ok(implement_kbd(
         keyboard,
         state,
-        key_repeat_kind,
-        Arc::new(Mutex::new(implementation)),
+        implementation,
     ))
 }
 
@@ -518,7 +509,6 @@ where
 pub fn map_keyboard_rmlvo<Impl>(
     keyboard: NewProxy<wl_keyboard::WlKeyboard>,
     rmlvo: RMLVO,
-    key_repeat_kind: KeyRepeatKind,
     implementation: Impl,
 ) -> Result<Proxy<wl_keyboard::WlKeyboard>, (Error, NewProxy<wl_keyboard::WlKeyboard>)>
 where
@@ -558,8 +548,7 @@ where
         Ok(state) => Ok(implement_kbd(
             keyboard,
             state,
-            key_repeat_kind,
-            Arc::new(Mutex::new(implementation)),
+            implementation,
         )),
         Err(error) => return Err((error, keyboard)),
     }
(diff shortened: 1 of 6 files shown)


trimental:
Doesn't build on 1.20.0  but I plan to change the way repeatable characters are chosen anyway

trimental:
Alacrity works but only displays the characters as you hold down the key if you also simultaneously hovering your mouse over the buttons which also means that the hover colors are properly displayed as well. Very strange! Maybe alacrity, glutin or winit suppress window drawing which would explain the buttons? or maybe a bug.

vberger:
Hmm, I'm a little uneasy about integrating it like that. While I agree it would be good to integrate key repetitions, I feel like implementing it using threads is pretty eavy, especially if it means spawning a new thread every key press.

Also, what you encounter here:

> Alacrity works but only displays the characters as you hold down the key if you also simultaneously hovering your mouse over the buttons which also means that the hover colors are properly displayed as well.

Is likely because the winit event loop only waits on events from the wayland connection, and not on events from your repetition thread. As such, if you don't move the pointer, no wayland events are generated, and the event loop remains stuck.

Because of this, I don't think it'll be possible to implement key repetition in a way that "just works", without special  integration in the event loop (by making it possible to wait on more event sources than just the wayland socket), and thus cooperation from the downstream user.

trimental:
I also think the current implementation is crude at present but I am unsure what you mean by not using threads, I think to do something like this you need to use threads either directly or indirectly. As for the threads spawning every key press I can make it so only one thread exist for the key repetition and thread sleeping can be replaced by tick from the chan crate. That being said if you want it done without threads directly I can use something like tokio. 👍

vberger:
In all cases, independently of the actual way it is implemented:

- there should be a flag at keyboard creation for the user to request what kind of key repetition they want, probably something like:

```rust
pub KeyRepeatKind {
    None, // no key repeat
    Fixed { rate: u64, delay: u64 }, // use a fixed rate/delay configuration
    System, // use the rate/delay configuration provided by the server
}
```

- the key events should contain something allowing the user to differentiate repetition event from actual keypresses

That said, sorry I had glanced too quickly at your code yesterday, I see you are directly calling the callbacks from the secondary thread.

By my other message still explains your weird bug though: winit actually buffers the events generated in the `event_loop.dispatch()` call, and only then sends them to the user callback. Meaning it still only blocks on the wayland socket, and the key repetition events will only be properly processed only if other wayland events are generated as well (by example with the user moving the pointer).

I'm not completely clear about what would be the best way to tackle this cleanly, though.

trimental:
I have made the chan crate handle the timing and interaction between the main and secondary thread, this should now mean the secondary thread is immediately closed when not needed. 
Technically you were correct, every key press spawns a new thread that controls the sending of that key's events however its designed so that only one thread can exist at any time. If you wanted to you could have one thread made at the start of the keyboard setup and then send the current key through a channel to that thread but unless the actual starting of a thread has a higher resource cost then keeping one open even when not needed then I think its better to start one for every key.

trimental:
Hopefully this should be pretty near complete. I have two questions about it though, would it be better to use rate and delay to increment the time member on the key event when it's sent or set it to none or something, I ask because I don't really know how time is used or how it's value changes. Secondly, chan 1.21 depends on rand 0.3 which can compile on rustc 1.20, however I am pretty sure cargo updates it to rand 0.5 which can only compile on rustc 1.22., is there a way to stop this upgrade?

trimental:
Also does,

> I'm not completely clear about what would be the best way to tackle this cleanly, though.

refer to the client-toolkit implementation or the winit intergration

trimental:
@heghe crossbeam would be great to replace chan here :) but unfortunately it’s minimum supported rust version is 1.26 and this project has a minimum of 1.20. Technically this is a breaking change pull request so the minimum supported version could be bumped but I assume @vberger wouldn’t want to do that.

trimental:
Mm I’m starting to question if the chan/crossbeam crates are either A - important for function or B - performance efficient when compared to the original thread sleep way. With respect to A it should function the same if a thread kill check is done at least before the sending of an event and with respect to B it should be just as fast if not faster maybe. I think the chan crate might just be adding a nice front end to the original way.

trimental:
Looks like it's tempfile that is actually messing up the build process, they upgraded rand to 5 and released it as a compatible upgrade. The fix would be bumping the minimum supported rust version to 1.22 or removing tempfile. Regardless of this I would still prefer to not use the chan crate, I don't think it provides a lot for this usage scenario.

trimental:
Looks like it's tempfile that is actually messing up the build process, they upgraded rand to 5 and released it as a non-major release. The fix would be bumping the minimum supported rust version to 1.22 or removing tempfile. Regardless of this I would still prefer to not use the chan crate, I don't think it provides a lot for this usage scenario.

vberger:
Okay. Sorry for the delay, I had quite a lot going on and only now can I take the time to sit down and check your PR properly.

Overall the side-thread implementation you have settled on seems fine, but I have some concerns API-wise:

- First, could you explain to me some more the `RepeatedKeyTypes` option? Is there really such a strong need for this that it needs to be put in SCTK, rather than letting the user filter the repeat events as they see fit? To be honest I don't see an example of an app that would be interested in repeat keys, but only from some keys...
- Second, my previous concern about event loop integration still stands. Up to now, all event callbacks have the property that they will always be called during `event_queue.dispatch()`. Your PR here changes that, making the callback for key events be called independently. This may cause unexpected behavior if the app using SCTK does not take that into account, with basically the app blocking waiting for a wayland event, and when the user moves their mouse or release the key, all repeat events end up processed at once, rather than being processed every few millisecond as they should.

To circumvent that, I think things would be overall clearer if the user provided a second callback, handling only key repetition.

So, this would separate things much more. The keyboard event enum would not change, and rather a new `KeyRepeatEvent` struct could be added. The various `map_keyboard..` functions would take one more optional argument, or even they could be separated into new constructor methods `map_keyboard_auto_with_repeat(..)`, `map_keyboard_rmlvo_with_repeat(..)` (or whatever better name can be found) that take additional arguments: the repeat callback and the repetition parameters. And document clearly that this second callback can be fired anytime, not only during `event_queue.dispatch()`.

I believe that organizing the API like this would make things easier to handle for downstream crates. What do you think?

trimental:
The RepeatedKeyTypes setting was just for a bit more fine tuning however I completely agree that this could be easily implemented by the user instead.
Maybe map_keyboard could have a second implementation argument that's wrapped in a option? so disabling key repetition would be like
``` 
map_keyboard(bar, foo, |e| { handle key event }, |e| { handle repeat event }); // for enable
map_keyboard(bar, foo, |e| { handle key event }, None); // for disable
```
Btw huge congratulations on the wayland rs change, I saw the post on reddit, viva la pure rust revolution

trimental:
Just checking, are we also repeating characters like ctrl and shift and letting the user deal with that or only characters that are traditionally repeatable like backspace, tab and visible characters?

vberger:
After checking, we should use the [`xkb_keymap_key_repeats`](https://xkbcommon.org/doc/current/group__components.html#ga9d7f998efeca98b3afc7c257bbac90a8) from libxkbcommon.

It is not yet in the ffi module, so should be added here: https://github.com/Smithay/client-toolkit/blob/master/src/keyboard/ffi.rs#L177

trimental:
I made a few changes that separate the functions to map_keyboard and map_keyboard_with_repetition. I am unsure whether the closure passed for repeat callback needs to be static or if there's another way to implement it (there probably is but I'm not great at lifetimes).

trimental:
I think that with these last commits I've completed the changes to API you've requested and fixed the building on all rust versions. Please tell me if you want the API reworked a little bit more and I'll go over the code a few times to make sure that its working as expected.

src/keyboard/mod.rs
```diff
@@ -356,6 +352,22 @@ impl Drop for KbState {
     }
 }
 
+/// Determines the behaviour of key repetition
+#[derive(PartialEq)]
+pub enum KeyRepeatKind {
+    /// keys will not be repeated
+    None,
```
vberger: Given we have split the repeat handling into a new function, this variant is not necessary any more.

src/keyboard/mod.rs
```diff
@@ -435,6 +447,26 @@ pub enum Event<'a> {
...
+    /// interpreted symbol of the key
+    pub keysym: u32,
+    /// new state of the key
+    pub state: KeyState,
+    /// utf8 interpretation of the entered text
+    ///
+    /// will always be `None` on key release events
+    pub utf8: Option<String>,
+}
+
```
vberger: Some fields can be removed from here:

- `serial` is only meaningful for key press events, not event we generate ourselves such as these repeat events
- `state`: as this is a repeat event, this will always be `Pressed`, so it does not convey any useful information

src/keyboard/mod.rs
```diff
@@ -625,3 +657,311 @@ where
         },
     )
 }
+
+/// Implement a keyboard to automatically detect the keymap and send KeyRepeatEvents
+/// at set intervals
+///
+/// This requires you to provide an implementation to receive the events after they
+/// have been interpreted with the keymap, as well as an implementation to be called
+/// when KeyRepeatEvents are sent at intervals set by the KeyRepeatKind argument.
```
vberger: Please add a note like "This second implementation can be called anytime independently of the dispatching of the wayland events."

src/keyboard/mod.rs
```diff
@@ -625,3 +657,311 @@ where
...
+        Arc::new(Mutex::new(repeat_implementation)),
+    ))
+}
+
+/// Implement a keyboard for a predefined keymap and send KeyRepeatEvents at set
+/// intervals
+///
+/// This requires you to provide an implementation to receive the events after they
+/// have been interpreted with the keymap, as well as an implementation to be called
+/// when KeyRepeatEvents are sent at intervals set by the KeyRepeatKind argument.
```
vberger: Same note here about dispatching.

src/keyboard/mod.rs
```diff
@@ -625,3 +657,311 @@ where
...
+/// Returns an error if xkbcommon could not be initialized.
+pub fn map_keyboard_auto_with_repeat<Impl, RepeatImpl>(
+    keyboard: NewProxy<wl_keyboard::WlKeyboard>,
+    key_repeat_kind: KeyRepeatKind,
+    implementation: Impl,
+    repeat_implementation: RepeatImpl,
+) -> Result<Proxy<wl_keyboard::WlKeyboard>, (Error, NewProxy<wl_keyboard::WlKeyboard>)>
+where
+    for<'a> Impl: Implementation<Proxy<wl_keyboard::WlKeyboard>, Event<'a>> + Send,
+    RepeatImpl: Fn(KeyRepeatEvent) + Send + 'static,
```
vberger: Please use the `Implementation` trait here as well, so

`RepeatImpl: Implementation<Proxy<wl_keyboard::WlKeyboard>, KeyRepeatEvent> + Send`

src/keyboard/mod.rs
```diff
@@ -625,3 +657,311 @@ where
...
+pub fn map_keyboard_rmlvo_with_repeat<Impl, RepeatImpl>(
+    keyboard: NewProxy<wl_keyboard::WlKeyboard>,
+    rmlvo: RMLVO,
+    key_repeat_kind: KeyRepeatKind,
+    implementation: Impl,
+    repeat_implementation: RepeatImpl,
+) -> Result<Proxy<wl_keyboard::WlKeyboard>, (Error, NewProxy<wl_keyboard::WlKeyboard>)>
+where
+    for<'a> Impl: Implementation<Proxy<wl_keyboard::WlKeyboard>, Event<'a>> + Send,
+    RepeatImpl: Fn(KeyRepeatEvent) + Send + 'static,
```
vberger: Same here for `RepeatImpl`

src/keyboard/mod.rs
```diff
@@ -625,3 +657,311 @@ where
...
+fn implement_kbd_with_repeat<Impl, RepeatImpl>(
+    kbd: NewProxy<wl_keyboard::WlKeyboard>,
+    mut state: KbState,
+    key_repeat_kind: KeyRepeatKind,
+    mut user_impl: Impl,
+    repeat_impl: Arc<Mutex<RepeatImpl>>,
+) -> Proxy<wl_keyboard::WlKeyboard>
+where
+    for<'a> Impl: Implementation<Proxy<wl_keyboard::WlKeyboard>, Event<'a>> + Send,
+    RepeatImpl: Fn(KeyRepeatEvent) + Send + 'static,
```
vberger: Same here for `RepeatImpl`

src/keyboard/mod.rs
```diff
@@ -625,3 +657,311 @@ where
...
+                                            modifiers,
+                                            rawkey: key,
+                                            keysym: sym,
+                                            state: key_state,
+                                            utf8: utf8.clone(),
+                                        },
+                                        proxy,
+                                    );
+                                    // Start thread to send key events
+                                    thread::spawn(move || {
```
vberger: So, here you spawn a new thread on keypress.

How does it behave if two repeating keys are pressed simultaneously?

Like, if I press `a`, it starts repeating. Then I press `z` without releasing `a`, what happens?

What my current DE does is that the previously repeating key stops repeating, and the new one is now the only one repeating.

trimental: My previous logic for this was wrong (I did have it at one time but must of removed it). Now it checks that no other key is being held before it sends any new key events. If you want I can change this so it sends key events but doesn't repeat them if another key is being pressed instead.

src/keyboard/mod.rs
trimental: I've changed it to try and explain it in a way that made sense to me and flowed with what was already there but if you can think of a clearer way to say it then I'll change it to that.

src/keyboard/mod.rs
trimental: Or I can make it can stop all key repetition (kill all threads) or kill the previous key repetition and just repeat the new key, its up to you.

vberger: Well, the behavior I've seen everywhere is that only the last pressed key repeats, so I'd say the best is probably to do the same thing.

trimental:
The reflow key repetition logic commit changes the behavior of key repetition to this
* All key events are guaranteed to send a key press and key release event
* Only the last pressed repeatable key sends repeat events
Due to these guarantees, if you ignore key repeat events then map_keyboard_with_repeat will essentially be map_keyboard. Press events of key A will never occur while repeating key B, release of events of key A may occur when repeating key B. I think this is the best behavior for key repetition

trimental:
I think I need some more clarification on 
> struct implementing the appropriate Implementation<...> trait.

do you mean a alternative to kbd.implement() like repeat_kbd.implement()?

src/keyboard/mod.rs
trimental: Nah this should hopefully only require an small adjustment to make state be passed by arc instead of pure clone.

vberger:
> Press events of key A will never occur while repeating key B, release of events of key A may occur when repeating key B.

Hmm, I'm not sure I understand what you mean here. I'll put an example of what I think is the expected behavior:

- user presses `a`
  - a key press event for `a` is generated
  - `a` starts repeating
- user presses `<shift>`
  - `a` stops repeating
  - a key press event for `<shift>` is generated
  - `A` starts repeating
- user releases `<shift>`
  - `A` stops repeating
  - a release event for `<shift>` is generated
  - `a` starts repeating
- user presses `b`
  - `a` stops repeating
  - a press event for `b` is generated
  - `b` starts repeating
- user releases `b`
  - `b`stops repeating
  - a release event for `b` is generated
- user releases `a`
  - a release event for `a` is generated

Do we agree on this stream of events?

> do you mean a alternative to kbd.implement() like repeat_kbd.implement()?

I mean, when implementing the `NewProxy<WlKeyboard>`, rather than giving this huge closure to the `implement()` method, instead add a new struct `KbdImplementation` which implements the trait `Implementation<Proxy<WlKeyboard>, wl_keyboard::Event>`:

```rust
impl Implementation<Proxy<WlKeyboard>, wl_keyboard::Event> for KbdImplementation {
    fn receive(&mut self, event: wl_keyboard::Event, kbd: Proxy<WlKeyboard>) {
        /* contents of the big closure */
    }
}
```

This would allow to more easily factor code between the two implementations, and possibly split the whole logic into a few methods of `KbdImplementation` for improved readability.

trimental:
I'm sorry I'm not that use to the Implement trait, I got something like this
```
struct KbdImplementation {
    state: KbState,
    key_repeat_kind: KeyRepeatKind,
    kill_chan: Arc<Mutex<(mpsc::Sender<()>, mpsc::Receiver<()>)>>,
    key_held: Option<u32>,
    system_repeat_timing: Arc<Mutex<(u64, u64)>>,
}

impl Implementation<Proxy<wl_keyboard::WlKeyboard>, wl_keyboard::Event> for KbdImplementation {
    fn receive(&mut self, event: wl_keyboard::Event, kbd: Proxy<wl_keyboard::WlKeyboard>) {
// Big closure matching event
   }
}
```
does KbdClosure need to take user_impl and repeat_impl?

trimental:
I might have to complete this pull request without this and leave you to implement it later if I can't figure it out from your explanations.

trimental:
Unless I'm mistaken you could also maybe have this
```
fn implement_kbd<Impl, RepeatImpl>(
    kbd: NewProxy<wl_keyboard::WlKeyboard>,
    mut state: KbState,
    mut user_impl: Impl,
    repeat: Option<(RepeatImpl, KeyRepeatKind)> 
) -> Proxy<wl_keyboard::WlKeyboard>
```

map_keyboard would pass none and map_keyboard_with_repeat would pass values. Then it would be as simple as a if let in the Key Press arm of the event match. Your way might be better though as I don't fully understand it yet.

vberger:
The idea of the 'Implementation' trait is more or less the concretisation of the fact that a closure if really just a struct that happens to be callable.

As such, if you want to transform a closure into a struct, you indeed have to create a struct whose fields are exactly the variables that the closure captured. Then, in the implementation, just use the same code as what the closure had, replacing the captured variable by access to the fields.

When doing this kind of migration, I just tend to move the code from the closure to the impl block then follow the compiler error messages until it compiles. ^^"

As an example, you can see some direct implementations in smithay, here for example: https://github.com/Smithay/smithay/blob/master/src/wayland/shm/mod.rs

vberger:
Indeed, having an Option argument would work too, if you prefer that way. :)

trimental:
When I did
> move the code from the closure to the impl block then follow the compiler error messages until it compiles."

I was left with a final error message saying this for the Impl as well as RepeatImpl
> wayland_commons::Implementation<wayland_client::Proxy<wayland_client::protocol::wl_keyboard::WlKeyboard>, keyboard::Event<'a>> + std::marker::Send + 'static` does not have a constant size known at compile-time

I'll write it as a Option for now but you can change it to a struct Implementation if you can get it working

trimental:
Also I forgot to answer
> Hmm, I'm not sure I understand what you mean here. I'll put an example of what I think is the expected behavior:

Yes that should be the behavior

src/keyboard/mod.rs
```diff
@@ -569,59 +604,267 @@ where
...
                 wl_keyboard::Event::Modifiers {
                     mods_depressed,
                     mods_latched,
                     mods_locked,
                     group,
                     ..
-                } => state.update_modifiers(mods_depressed, mods_latched, mods_locked, group),
+                } => {
+                    state.lock().unwrap().update_modifiers(mods_depressed, mods_latched, mods_locked, group);
+                    state_chan.lock().unwrap().0.send(()).unwrap();
```
vberger: Doesn't that stop any current repetition if one was going on?
Pressing a modifier should change the repeated key but not stop the repetition.

Also, what's the impact of sending a message in this channel if no repetition is going on?

src/keyboard/mod.rs
```diff
@@ -569,59 +604,267 @@ where
...
+/// This requires you to provide an implementation to receive the events after they
+/// have been interpreted with the keymap. You must also provide an implementation to be called
+/// when KeyRepeatEvents are sent at intervals set by the KeyRepeatKind argument, this
+/// implementation can be called at anytime, independent of the dispatching of wayland events.
+///
+/// The keymap information will be loaded from the events sent by the compositor,
+/// as such you need to call this method as soon as you have created the keyboard
+/// to make sure this event does not get lost.
+///
+/// Returns an error if xkbcommon could not be initialized.
```
vberger: Please add a note in this documentation and the one of `map_keyboard_rmlvo_with_repeat` that threads will be spawned in the background. I prefer this to be explicit.

src/keyboard/mod.rs
```diff
@@ -569,59 +612,252 @@ where
...
+                                rawkey: key,
+                                keysym: sym,
+                                state: key_state,
+                                utf8: utf8.clone(),
+                            },
+                            proxy.clone(),
+                        );
+                        if let Some(repeat_impl) = repeat_impl.clone() {
+                            // Check with xkb if key is repeatable
+                            if unsafe { state.key_repeats(key + 8) } {
```
vberger: the `+ 8` offsetting should happen inside of the `key_repeats` method, like for the other methods of `KbState`.

src/keyboard/mod.rs
```diff
@@ -569,59 +612,252 @@ where
...
                     mods_latched,
                     mods_locked,
                     group,
                     ..
-                } => state.update_modifiers(mods_depressed, mods_latched, mods_locked, group),
+                } => {
+                    state.update_modifiers(mods_depressed, mods_latched, mods_locked, group);
+                    if key_held.is_some() {
+                        state_chan.lock().unwrap().0.send(()).unwrap();
+                    }
```
vberger: This code still means that pressing a modifier will stop any repeat, right? I don't think that's the expected behavior.

If I hold `a` on my computer and press then release `<shift>` several times, I get something like `aaaaaaaaaaaaaAAAAAAAAAAaaaaaaaaaaAAAAAAaaaaaaaaaAAAAAaaaaaaaa`.

trimental: Maybe you are thinking of **kill**_chan instead of **state**_chan, state_chan simply requests the thread to update modifiers when they are changed rather then updating the values every 30ms. I don't really know whether this is much more performant then just updating them for every key repetition though.

trimental: Actually re-reading that, is 
> aaaaaaaaaaaaaAAAAAAAAAAaaaaaaaaaaAAAAAAaaaaaaaaaAAAAAaaaaaaaa

not the expected behavior? what would be?

vberger: > Maybe you are thinking of kill_chan instead of state_chan

Ah right, my bad. I've read it too quickly. No problem then.

trimental:
I've tried to implement all of your review points, the only one I had difficulty with was
> I think it'd be better if the wrapping in Arc<Mutex<...>> was done inside of this function

as I can't move RepeatImpl out of a container as it needs a size, unless I move it to generics, which I can't do because then all calls must provide a RepeatImpl object, when some pass None. (atleast I'm pretty sure thats why)

> Sorry for the long back and forth on this PR, thanks for enduring me this whole time

Not at all, I love learning more about the wayland protocol and your reviews have been spot on, its best to get these things as stable as possible :)

trimental:
Works beautifully, the parameters and their calling looking a lot better then before

PR #17, opened 2018-07-21
Bump minimum rust version

trimental:
The update of tempfile 3.0.3 bumps rand to 0.5 which means that the client toolkit won't compile on rust 1.20. This pull request shows that the minimum rust version needed to compile the client toolkit is rust 1.22.


vberger:
Alright. I'm not fan of bumping this requirement, but looks like we don't really have a choice.

PR #19, opened 2018-07-25
More linux style buttons

trimental:
screenshot from 2018-07-25 08-36-48
This one doesn't really need a whole lot of explanation, it adds a new style of buttons that mimics the style of linux decorations. Although this isn't the only 'popular' button style on linux, I think it is the most popular as it's gnomes default style and therefore fedoras and ubuntus..ect.

### DIFF
--- a/src/window/basic_frame.rs
+++ b/src/window/basic_frame.rs
@@ -42,6 +42,8 @@ mod colors {
     pub const YELLOW_BUTTON_REGULAR: &[u8] = &[0x40, 0xB0, 0xB0, 0xFF];
     pub const YELLOW_BUTTON_HOVER: &[u8] = &[0x40, 0xFF, 0xFF, 0xFF];
     pub const YELLOW_BUTTON_DISABLED: &[u8] = &[0x20, 0x80, 0x80, 0xFF];
+    pub const BUTTON_OUTLINE: &[u8] = &[0x90, 0x90, 0x90, 0xFF];
+    pub const BUTTON_HOVER_SHADE: &[u8] = &[0x8a, 0x8a, 0x8a, 0xFF];
 }
 #[cfg(target_endian = "big")]
 mod colors {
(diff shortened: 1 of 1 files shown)


vberger:
I not very fond of the idea of mimicking the default GTK theme. My worry is that if SCTK's decorations start looking too much like them, people might start getting confused when they aren't affected by GTK's configuration.

For thir reason, I'm thinking it'd better if SCTK had its own decoration style.

trimental:
I would say that changing the maximize button to be a bigger square would be the way to make this difference clear. The minimize and close button are pretty standard and generic (ubuntu and windows), I don't think it would be easy to make these much more different. Do you agree that a change to the maximize button would be sufficient? a small maximize box seams to be the biggest indicator of gtk buttons.

trimental:
Maybe also a readjustment of the space between buttons

trimental:
screenshot from 2018-07-25 16-07-39
Heres what that would look like, the white border is to show the buttons better as they are colored black

trimental:
I'd like to explain more that my decision to implement the 'line, box, cross' buttons seen in windows and gnome rather then the arrow buttons seen on KDE or the colored circles on MacOS, etc, is more about simplicity then GTK mimicry. The 'line, box, cross' buttons are minimalist and instantly recognizable. While without gtk and qt theme support you can never achieve the user's ideal look, I would say these buttons are a sensible default. Obviously though I am completely open to changing the look of the buttons to whatever is decided to be the best design.

vberger:
> I would say that changing the maximize button to be a bigger square would be the way to make this difference clear. The minimize and close button are pretty standard and generic (ubuntu and windows), I don't think it would be easy to make these much more different.

On the other hand, look at KDE, XFCE, MacOS, or even [the weston apps](https://linux-cdn.softpedia.com/screenshots/Wayland_1.png)... There is much more variety. Even on Gnome, if the user bothers to change their theme.

- First the buttons can be actually materialized (with rectangles, like weston and some XFCE themes, or circles, like on macOS or some KDE themes)
- Then, there is the possibility to use other symbols. For example some decorations use a `+` or a `^` for the maximize button. Others use `v` for the minimize button.
- Also, the buttons can be colored. The close button is for example often red.

Really, I'm not fan of "just mimicking the default GTK theme". Or if we try to be GTK-compatible, then we should parse the GTK configuration and draw like the actual GTK theme of the user.

vberger:
My main point is, I don't want final users to erroneously believe the app they are using is a GTK app because it kinda looks like the default GTK theme, and them come complaining to the app writer that the app does not respect GTK's configuration. This would just be frustrating and confusing for everyone.

vberger:
> the idea of creating a separate file containing pure drawing methods (lines, shaded boxes, circles, gradients) to assist in decorating the window?

This would definitely be helpful, but I believe it should probably be a separate crate, like something that would take a `&mut [u8]` as input, which would be a view into the mem-mapped buffer, and use it as the buffer of a canvas to draw on. This would certainly be useful not only in SCTK.

So if you feel like creating such a crate and port the decoration-drawing code to it, it'd certainly simplify the logic a lot, and possibly allow more easily to even provide choice between different `Frame` implementation.

trimental:
I'll try to create a good drawing crate for use in the drawing of the decorations. Fingers crossed it will turn out ok :)

vberger:
This is superseeded by #21 I believe

PR #20, opened 2018-07-27
Server decorations

### LINKED ISSUE #1: Integrate KDE's decoration protocol
An early version is available in wayland-protocols : https://docs.rs/wayland-protocols/0.20.1/wayland_protocols/misc/server_decoration/client/index.html

There is also ongoing discussion on the wayland mailling lists and some variant of the protocol is likely to be soon standardized : https://lists.freedesktop.org/archives/wayland-devel/2018-April/037933.html

The idea would be to make a `ServerFrame<F: Frame>` implementing `window::Frame` that would use the server decoration protocol if available and if not fallback to the `F` frame.

vberger:
fixes #1 

It only work with the `xdg-decorations`, which works with the stage `xdg-shell` protocol.

Not sure if there is already a compositor implementing both for testing...

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -82,7 +82,8 @@ struct WindowInner<F> {
 pub struct Window<F: Frame> {
     frame: Arc<Mutex<F>>,
     surface: Proxy<wl_surface::WlSurface>,
-    decoration: Option<Proxy<zxdg_toplevel_decoration_v1::ZxdgToplevelDecorationV1>>,
+    decoration: Mutex<Option<Proxy<zxdg_toplevel_decoration_v1::ZxdgToplevelDecorationV1>>>,
+    decoration_mgr: Option<Proxy<zxdg_decoration_manager_v1::ZxdgDecorationManagerV1>>,
     shell_surface: Arc<Box<shell::ShellSurface>>,
     inner: Arc<Mutex<Option<WindowInner<F>>>>,
 }
@@ -116,8 +117,8 @@ impl<F: Frame + 'static> Window<F> {
         )
     }
 
-    /// Create a new window wrapping a given wayland surface as its main content and using
-    /// server decorations if available
+    /// Create a new window wrapping a given wayland surface as its main content and
+    /// following the compositor's preference regarding server-side decorations
     ///
     /// It can fail if the initialization of the frame fails (for example if the
     /// frame class fails to initialize its SHM).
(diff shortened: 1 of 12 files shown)


emersion:
Tested with https://github.com/swaywm/wlroots/pull/1053, seems to work well! To change the compositor's preference, change `WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_CLIENT_SIDE` in `rootston/xdg_shell.c` line 491 to `WLR_XDG_TOPLEVEL_DECORATION_V1_MODE_SERVER_SIDE`.

src/window/mod.rs
```diff
@@ -92,6 +101,36 @@ impl<F: Frame + 'static> Window<F> {
...
+            subcompositor,
+            shm,
+            shell,
+            None,
+            implementation,
+        )
+    }
+
+    /// Create a new window wrapping a given wayland surface as its main content and using
+    /// server decorations if available
```
emersion: You can set the client's preference to SSD if you prefer to use SSD. Right now the client preference is not set, so the compositor decides.

src/window/mod.rs
```diff
@@ -253,7 +324,18 @@ impl<F: Frame + 'static> Window<F> {
...
+        use self::zxdg_toplevel_decoration_v1::Mode;
         self.frame.lock().unwrap().set_hidden(!decorate);
+        if let Some(ref dec) = self.decoration {
+            if decorate {
+                // let the server decide decorations
+                dec.unset_mode();
+            } else {
+                // we must be the decorators, to be able to not
+                // show any decoration
+                dec.set_mode(Mode::ClientSide);
```
emersion: It might be more complicated to implement, but if you don't want to be decorated, you should destroy the decoration object. The compositor always has the last word wrt decoration mode.

src/window/mod.rs
vberger: My goal with that was mostly to aim for "follow the compositor's preference". But yeah, I need to clarify what policy to actually follow.

src/window/mod.rs
vberger: Right, indeed. This will be more convoluted to implement though.

newpavlov:
Gnome on Wayland officially [will not support](https://gitlab.gnome.org/GNOME/mutter/issues/217) xdg-decorations.

vberger:
@newpavlov Do you mean this as an opinion against implementing it in SCTK?

My approach is to target maximal compatibility. Gnome has its opinion on how clients should behave and look like, but Gnome is also not the only player in the field. If a compositor supports SSD, I see no downside on using them, especially from the point of view of the users of SCTK-based apps.

newpavlov:
No, it's just a notice for those who hope to use SSD on Gnome.

PR #21, opened 2018-07-29
Draw decorations using a seperate crate

trimental:
I've done some work on moving the decoration drawing code to a new crate called andrew (the name is a play on 'drew' being the past tense of 'draw'). Currently the crate is quite basic with its only functionality being bresenham lines and bordered/rounded rectangles or squares. I hope to add some more geometric objects and effects before the 0.1 release, most important being circles, circular sectors, shadows and gradients. 

The current fork doesn't work yet as andrew isn't released but I can tell you that it looks the exact same as before. The main purpose of this pull request is so that the crate's api can be reviewed and reworked while its still in a non-release stage. I'll give a little explanation on the crates api so far.

The Draw trait which looks like this
```
pub trait Draw {
    fn draw(&self, canvas: &mut [u8], canvas_size: (usize, usize));
}
```
allows an object to be drawn to the screen. This trait has many benefits but one of the biggest I can think of is the ability to have whole decorations contained in one object. 

The crate is currently split into two modules, line and shape. The only public object that the line module contains at the moment is the Line struct that can be drawn with the syntax of
```
Line::new( pt1: (usize, usize), pt2: (usize, usize) ).draw(canvas: &mut [u8], canvas_size: (usize, usize))
```
this syntax will probably be changed in the future to support antialising.

The shape module currently only contains the rectangle submodule, which contains the main Rectangle struct as well a bitflag Sides struct with the flags of TOP, BOTTOM, LEFT, RIGHT, ALL. The api for drawing a rectangle looks like this
```
Rectangle {
    position of rectangle: (usize, usize),
    size of rectangle: (usize, usize),
    a option for a border with a width, color, side and roundness: Option<(usize, [u8; 4], Sides, Option<usize>)>,
    a option for a colored fill: Option<[u8; 4]>,
}
```
The Sides struct allows only parts of the rectangle to be bordered. Its worth mentioning that colors passed to andrew are automatically reversed (only once, so minuscule runtime cost) when on a little endian machine, this means that all colors that are passed are in the [A, R, G, B] format. 

I would like feedback on the api of the crate such as whether the drawing of objects should be done through a struct + trait or just functions and whether co-orientates and sizes should be stored as usizes or u32s like in the client-toolkit. I would also be happy to change the name to something more technical like BufDraw or something if you'd like :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- MemPool: add `mmap` method
 - **[Breaking]** Keyboard: remove `modifiers` field from `keyboard::Event::Enter`, `keyboard::Event::Key` and `keyboard::KeyRepeatEvent`
 - **[Breaking]** Keyboard: add `keyboard::Event::Modifiers`
 - **[Breaking]** Upgrade to wayland-rs 0.21
--- a/src/utils/mempool.rs
+++ b/src/utils/mempool.rs
@@ -12,7 +12,7 @@ use std::os::unix::io::FromRawFd;
 use std::os::unix::io::RawFd;
 use std::sync::{Arc, Mutex};
 
-use memmap;
+use memmap::MmapMut;
 use rand::prelude::*;
 
 use wayland_client::protocol::{wl_buffer, wl_shm, wl_shm_pool};
@@ -104,6 +104,7 @@ pub struct MemPool {
     len: usize,
     pool: Proxy<wl_shm_pool::WlShmPool>,
     buffer_count: Arc<Mutex<u32>>,
+    mmap: MmapMut,
     implementation: Arc<Mutex<FnMut() + Send>>,
 }
 
@@ -121,11 +122,14 @@ impl MemPool {
             .create_pool(mem_fd, 128, |pool| pool.implement(|e, _| match e {}, ()))
             .unwrap();
 
+        let mmap = unsafe { MmapMut::map_mut(&mem_file).unwrap() };
+
         Ok(MemPool {
             file: mem_file,
             len: 128,
             pool,
             buffer_count: Arc::new(Mutex::new(0)),
+            mmap,
             implementation: Arc::new(Mutex::new(implementation)),
         })
     }
@@ -146,6 +150,7 @@ impl MemPool {
             self.file.set_len(newsize as u64)?;
             self.pool.resize(newsize as i32);
             self.len = newsize;
+            self.mmap = unsafe { MmapMut::map_mut(&self.file).unwrap() };
         }
         Ok(())
     }
@@ -192,8 +197,8 @@ impl MemPool {
     }
 
     /// Uses the memmap crate to map the underlying shared memory file
-    pub fn mmap(&self) -> Result<memmap::MmapMut, io::Error> {
-        unsafe { memmap::MmapMut::map_mut(&self.file) }
+    pub fn mmap(&mut self) -> &mut MmapMut {
+        &mut self.mmap
     }
 
     /// Retuns true if the pool contains buffers that are currently in use by the server otherwise it returns
--- a/src/utils/mempool.rs
+++ b/src/utils/mempool.rs
@@ -12,6 +12,7 @@ use std::os::unix::io::FromRawFd;
 use std::os::unix::io::RawFd;
 use std::sync::{Arc, Mutex};
 
+use memmap;
 use rand::prelude::*;
 
 use wayland_client::protocol::{wl_buffer, wl_shm, wl_shm_pool};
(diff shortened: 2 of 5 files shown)


vberger:
Given your crate is not yet I can't comment it directly, but I can tell you what I think of what you described so far. :)

Regarding the same: it's your crate, you get to name it however you like. Though I'll still say that `andrew` sounds nice to my ears :+1: .

Regarding your choice of trait for the API, I think the buffers should have a proper type representing them. My first instinct would be to have an API along the lines of:

```rust
pub trait Drawable {
    /* A trait for whatever can be drawn. Line, Rectangles, Circles, etc... */
}

pub struct Buffer {
    /* .... */
}

impl Canvas {
    /// Create a canvas from a mutable slice
    ///
    /// Here, width is the width in pixels of the drawing window, height is the
    /// height in pixels, and stride is the number of bytes separating the beginning
    /// of two consecutive lines.
    ///
    /// So, it needs to have the following invariants to be meaningful (and can panic if not met):
    /// - stride >= width * { number of bytes in a pixel }
    /// - stride * (height - 1) + width * { number of bytes in a pixel } <= buffer.len()
    fn new(buffer: &mut [u8], width: usize, height: usize, stride: usize) -> Canvas {
        /* .... */
    }

    /// Draw something on this canvas
    fn draw<D: Drawable>(&mut self, drawable: D) {
        /* .... */
    }
}
```

So, the user would do something like:

```rust
let mut canvas = Canvas::new(....);
canvas.draw(...);
canvas.draw(...);
```

So, this is already quite similar to your idea, just creating the canvas object. My reasoning for suggesting that is the following:

- It feels more natural and practical to me (this is highly subjective, of course)
- the `Drawable` trait becomes kind of a lower level API of your crate, this notably leaves room for adding methods in it for doing more efficient drawing of some shapes for example. Though I'm not really clear about whether there is actually something to gain here...

The handling or ARGB correctly independently of endianness is a nice touch also :)

Regarding u32/usize for drawing coordinates... If you want to be really generic and handle things like sub-pixel aliasing and such, well, they should be `f32` I'd say.

Also, I don't know how relevant it would actually be, but maybe the [`image`](https://docs.rs/image/0.19.0/image/) crate can have some interesting ideas to build upon? It's not really meant for drawing, rather low-level image manipulation and encoding, but it does not appear completely unrelevant either.

trimental:
I've just published 0.1.0 of [andrew](https://crates.io/crates/andrew) to crates.io. 

Right now there shouldn't be any major difference to the look of the decorations expect that the window rounding is changed due to the circle space (don't know what to call it) being rounded instead of just truncated.

I have not added the effects that I hoped would be in the 0.1.0 release (gradients, etc) as they are extremely dependent on the geometry of shapes and it would be hard to implement until the objects geometry stabilizes 

A feature I did get to add was the rendering of text which would allow the decorations to render the titles of applications, I think this would be a big improvement in the frame however I haven't added it yet as I would like to hear your opinion on it

I also removed the automatic ordering of bytes based on endianness as it essentially tethers andrew to wayland. Prehaps though a function or flag could be added to Canvas to turn on endianness specific drawing which would be set to little endian in this case.

src/utils/mempool.rs
```diff
@@ -190,6 +191,11 @@ impl MemPool {
             }).unwrap()
     }
 
+    /// Uses the memmap crate to map the underlying shared memory file
+    pub fn mmap(&self) -> Result<memmap::MmapMut, io::Error> {
+        unsafe { memmap::MmapMut::map_mut(&self.file) }
+    }
+
```
vberger: I'm wondering if it would be better to have the `MmapMut` be a field of `MemPool` and just have the method return a `&mut MmapMut`. This way, the lifetimes of the mmap and of the `MemPool` are tied. This just means the map must be destroyed and re-created every time the mempool size changes, but that's still less often than at every frame.

What do you think?

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- **[Breaking]** MemPool: add `mmap` method
```
vberger: I don't think adding a method is a breaking change :wink:

src/window/basic_frame.rs
```diff
@@ -413,56 +416,59 @@ impl Frame for BasicFrame {
...
 
-                // For every pixel in borders
-                for _ in BORDER_SIZE * width..pxcount {
-                    let _ = writer.write(&[0x00, 0x00, 0x00, 0x00]);
+                // For each pixel in borders
+                {
+                    let mut writer = &mut mmap[HEADER_SIZE as usize * width as usize * 4..];
+                    for _ in 0..pxcount {
+                        let _ = writer.write(&[0x00, 0x00, 0x00, 0x00]);
+                    }
```
vberger: hmm, if `writer` is of type `&mut [u8]`, I don't think this code will fill it with zeros, given the `io::Write` implementation for `&mut [u8]` cannot track advancement in the buffer. This likely just "works" because the default value for the content when growing a file is already 0.

If think this can be replaced by

```rust
for b in &mut mmap[HEADER_SIZE as usize * width as usize * 4..] {
    *b = 0x00;
}
```

CHANGELOG.md
trimental: I thought **[Breaking]** meant API breaking instead of application breaking, my bad :)

PR #23, opened 2018-08-01
Fix find_button to avoid subtraction overflow

trimental:
This pull request fixes subtraction overflow when calling find_button in basic_frame. This is a fatal issue as it crashes the program when the cursor is hovered over the button area when the window is too small for buttons to be drawn.

### DIFF
--- a/src/window/basic_frame.rs
+++ b/src/window/basic_frame.rs
@@ -183,21 +183,21 @@ fn precise_location(old: Location, width: u32, x: f64, y: f64) -> Location {
 }
 
 fn find_button(x: f64, y: f64, w: u32) -> Location {
-    if (w >= 24 + BUTTON_SPACE)
+    if (w >= 24 + 2 * BUTTON_SPACE)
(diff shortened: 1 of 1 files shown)


trimental:
The https://github.com/Smithay/client-toolkit/pull/23/commits/00cf19b3b74ca3f89673e997fd7e6f4987286b23 commit matches the drawing of the buttons with the find_button measurements. I changed the minimum distance from the left window edge to the button to be equal to BUTTON_SPACE, if you look at the previous calculation of *button_dist + 2 * ds(BORDER_SIZE)* you can see that it don't make sense anymore after I changed the position and size of the headerbar. 

I chose to set the minimum window-button distance before disappearing to BUTTON_SPACE instead of the original 0 as it looks more aesthetic and symmetrical when they disappear, the downside is that the buttons disappear slightly faster, aka the minimum window width for all three buttons is 108 instead of 98.

vberger:
Interesting corner case, thanks!

Issue #18, opened 2018-07-24
Shared memory anonymous files location


emersion:
I noticed the `wl_shm` helper uses a regular temporary directory to create an in-memory file. This should probably be replaced with `XDG_RUNTIME_DIR` as creating a file for `wl_shm` backed by a disk should really be avoided.

Also, it's possible to create the file, open it, and immediately unlink it to make sure we're not leaking it.

Bonus point: some systems support better ways to allocate an anonymous file (see https://github.com/myfreeweb/shmemfdrs). Some systems also support file sealing which can ensure the file won't shrink (shrinking a file can cause SIGBUS on the compositor side, it's currently mitigated with a signal handler but it's a pretty hacky solution).

vberger:
Hmm, for some reason I was persuaded that the `tempfile` crate created temporary files that were not actually linked to the disk. Rereading its documentation, it's actually not clear at all that this would be the case. Thanks for noticing this.

trimental:
I'm just wondering whether this is a discussion about the location of the memory files or a discussion on passing an anonymous file descriptor like [memfd](http://man7.org/linux/man-pages/man2/memfd_create.2.html), or both?

emersion:
The high priority issue is the temporary files location. Using `memfd` and sealing would be nice to have, but it's more in the wishlist than high priority IMHO.

vberger:
Actually @emersion 

> Also, it's possible to create the file, open it, and immediately unlink it to make sure we're not leaking it.

Isn't that what the `tempfile` file is actually doing already? see https://github.com/Stebalien/tempfile/blob/master/src/file/imp/unix.rs#L66-L72

emersion:
Oh right. It even uses `O_TMPFILE` instead when available (https://github.com/Stebalien/tempfile/blob/master/src/file/imp/unix.rs#L75).

vberger:
Done in #26

PR #24, opened 2018-08-01
Shared memory

trimental:
I think this addresses and fixes the issue https://github.com/Smithay/client-toolkit/issues/18 by storing the temp file in XDG_RUNTIME_DIR where possible.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,6 @@ nix = "0.11"
 dlib = "0.4"
 lazy_static = "1"
 memmap = "0.6"
-tempfile = "3.0"
 wayland-commons = { version = "0.20.11" }
 wayland-client = { version = "0.20.11", features = ["cursor"] }
 wayland-protocols = { version = "0.20.11", features = ["client", "unstable_protocols"] }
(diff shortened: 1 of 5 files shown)


vberger:
Given the comments in https://github.com/Smithay/client-toolkit/issues/18#issuecomment-409616269 , I think this PR is just re-implementing in SCTK what tempfile is actually already doing, so I don't think we actually need it.

trimental:
Yeah this is almost exactly what tempfile does, the only difference is it stores the wayland memory file in XDG_RUNTIME_DIR by default (I don't think tempfile does that), this may be technically a more 'correct' place for the file but I don't see any benefits apart from that? I just tried to implement what was discussed in https://github.com/Smithay/client-toolkit/issues/18, you can decide whether or not this has any really benefits and close this if not :)

vberger:
> in XDG_RUNTIME_DIR by default (I don't think tempfile does that), this may be technically a more 'correct' place for the file but I don't see any benefits apart from that?

I don't think this is relevant given the file does not actually exist in the filesystem anyway, as `XDG_RUNTIME_DIR` is for files that are to be shared by path with other processes.

PR #25, opened 2018-08-02
Only destroy buffers when not in use

trimental:
This pull request makes it so that buffers aren't destroyed when they are in use by the server. Inactive buffer proxies (buffers not in the current memory pool) are stored in a vec so that they can destroyed if another redraw is called before the buffers can be swapped. All active buffers must be released before a mempool swap can occur. Please note that this code is based on my understanding that release events will only be sent to buffers in the current memory pool.

### DIFF
--- a/examples/image_viewer.rs
+++ b/examples/image_viewer.rs
@@ -2,10 +2,8 @@ extern crate byteorder;
 extern crate image;
 extern crate smithay_client_toolkit as sctk;
 
-use std::cell::Cell;
 use std::env;
 use std::io::{BufWriter, Seek, SeekFrom, Write};
-use std::rc::Rc;
 use std::sync::{Arc, Mutex};
 
 use byteorder::{NativeEndian, WriteBytesExt};
@@ -14,8 +12,8 @@ use sctk::reexports::client::protocol::wl_buffer::RequestsTrait as BufferRequest
 use sctk::reexports::client::protocol::wl_compositor::RequestsTrait as CompositorRequests;
 use sctk::reexports::client::protocol::wl_display::RequestsTrait as DisplayRequests;
 use sctk::reexports::client::protocol::wl_surface::RequestsTrait as SurfaceRequests;
-use sctk::reexports::client::protocol::{wl_buffer, wl_seat, wl_shm};
-use sctk::reexports::client::{Display, NewProxy, Proxy};
+use sctk::reexports::client::protocol::{wl_buffer, wl_seat, wl_shm, wl_surface};
+use sctk::reexports::client::{Display, Proxy};
 use sctk::utils::{DoubleMemPool, MemPool};
 use sctk::window::{BasicFrame, Event as WEvent, State, Window};
 use sctk::Environment;
@@ -176,22 +174,13 @@ fn main() {
     // First, we initialize a few boolean flags that we'll use to track our state:
     // - the window needs to be redrawn
     let mut need_redraw = false;
-    // - the frame needs to be refreshed
-    let mut need_refresh = false;
-    // - have we done the original bootstrap? Depending on the underlying shell
-    //   protocol, we may need to wait until we receive a first configure event
-    //   before drawing our window. This flag checks if this is done.
-    let mut bootstrapped = false;
-    // - number of currently free memory pool, which will be shared by the buffer
-    //   implementations
-    let free_pools = Rc::new(Cell::new(2u32));
     // - are we currently in the process of being resized? (to draw the image or
     //   black content)
     let mut resizing = false;
-    // - a QueueToken, we'll need it to implement the buffers
-    let token = event_queue.get_token();
     // - the size of our contents
     let mut dimensions = image.dimensions();
+
+    let mut buffer = None;
 
     // if our shell does not need to ait for a configure event, we draw right away.
     //
@@ -202,8 +191,17 @@ fn main() {
     // But if we have fallbacked to wl_shell, we need to draw right away because we'll
     // never receive a configure event if we don't draw something...
     if !env.shell.needs_configure() {
-        need_redraw = true;
-        bootstrapped = true;
+        // initial draw to bootstrap on wl_shell
+        if let Some(pool) = pools.pool() {
+            redraw(
+                pool,
+                &mut buffer,
+                window.surface(),
+                dimensions,
+                if resizing { None } else { Some(&image) },
+            )
+        }
+        window.refresh();
     }
 
     // We can now actually enter the event loop!
@@ -217,7 +215,8 @@ fn main() {
             // We receive a Refresh event, store that we need to refresh the
             // frame
             Some(WEvent::Refresh) => {
-                need_refresh = true;
+                window.refresh();
+                window.surface().commit();
             }
             // We received a configure event, our action depends on its
             // contents
@@ -228,29 +227,19 @@ fn main() {
                 if let Some((w, h)) = new_size {
                     if dimensions != (w, h) {
                         dimensions = (w, h);
-                        need_redraw = true;
                     }
                 }
+                window.resize(dimensions.0, dimensions.1);
+                window.refresh();
                 // Are we currently resizing ?
                 // We check if a resizing just started or stopped,
                 // because in this case we'll swap between drawing black
                 // and drawing the window (or the reverse), and thus we need to
                 // redraw
                 let new_resizing = states.contains(&State::Resizing);
-                if new_resizing != resizing {
-                    need_redraw = true;
-                }
                 resizing = new_resizing;
 
-                // If we have not bootstrapped yet, this is the first configure
-                // event, time to draw for the first time.
-                if !bootstrapped {
-                    bootstrapped = true;
-                    need_redraw = true;
-                }
-
-                // In all cases, schedule a refresh
-                need_refresh = true;
+                need_redraw = true;
             }
             // No event, nothing new to do.
             None => {}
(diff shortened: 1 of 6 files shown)


trimental:
I've realized that this creates tears and glitches when resizing the window quickly, I'll try to find the issue.

trimental:
I've reworked the logic and fixed the screen tearing, I'll explain the process of double buffering as it is now. 

Firstly active_buffer_count is set to 0 and the inactive_buffers vector is created to store the proxies of buffers that aren't active (so that we can destroy them because they won't get release events). 

Then inactive_buffers is drained on each redraw so that only the most recently drawn buffers are stored in the inactive pool. 

Each created buffer is implemented so that if a server sends it a release event it will destroy its self, de-increment the active_buffer_count, and check if a pool swap can occur (if there are no active buffers and there are inactive buffers to swap to). 

Swapping the inactive buffers to active buffers is done by letting go of their handle in inactive_buffers (through .clear()) as the server will take care of destroying them (through release events), and setting the active buffer count to 5 (number of buffers).

When the buffers are freed but there are no inactive buffers to swap to, the swapping of the buffers will be done on the next redraw.

vberger:
I'm having a hard time following your intent in the way you handle double-buffering here. Mostly, I don't understand your notion of "active buffer" and "inactive buffer", could you explain in more detail what you are doing?

In my mind, the handling of double buffering was something like:

- for each pool, keep track of the not-yet-released buffers pointing to it
- once all buffers pointing to it have been released, a pool is marked as "free"
- when a draw is requested:
  - if there is a free pool, use it to draw (and it is then no longer free)
  - if not, defer the draw until a pool gets freed

I don't manage to map this mental model to your code. Notably your PR doe snot seem to wait until a pool is completely released before using it to draw.

trimental:
The 'active' buffers are buffers that are in the currently used pool and that have not been released. 'Inactive' buffers are buffers that are contained in the pool thats not in use. Perhaps queued_buffers would be a better name then inactive_buffers.

firstly when a redraw is called, all queued buffers (if any) are refreshed with the new buffers created on that redraw
```
let mut inactive_buffers = self.inactive_buffers.lock().unwrap();
for b in inactive_buffers.drain(..) {
     b.destroy();
}
...
inactive_buffers.push(foo_buffer)
```

Each buffer is implemented with the following closure, this closure will only be called on active buffers as they are freed by the server

```
wl_buffer::Event::Release => {
    buffer.destroy();
    *my_active_buffer_count.lock().unwrap() -= 1;
    if *my_active_buffer_count.lock().unwrap() == 0 && my_inactive_buffers.lock().unwrap().len() == 5 {
        my_pools.lock().unwrap().swap();
        my_inactive_buffers.lock().unwrap().clear();
        *my_active_buffer_count.lock().unwrap() = 5;
    }
}
```
the buffer upon receiving the release, destroys itself, de-increments the active buffer count (because it destroyed itself) and checks if it is the last active buffer to be destroyed, if it is then it checks if there are any queued (inactive) buffers to be swapped to and swaps if there are, when it swaps it clears the queued buffers as they are now 'active' and sets the active count to the number of buffers (5). This swapping handles any redraw called while the buffers were still in use.

This piece of code at the end of redraw checks if there are no active buffers and if so it sets the queued buffers it just drew to active.
```
if *self.active_buffer_count.lock().unwrap() == 0 {
    self.pools.lock().unwrap().swap();
    nactive_buffers.clear();
    *self.active_buffer_count.lock().unwrap() = 5;
}
```

I'm willing to implement this another way (through two vecs of both pools buffers or something) however I'm pretty sure this way also works.

trimental:
I’ve thought about it and I now think the tracking buffers and handling when to swap might be better abstracted through the double memory pool object

trimental:
Hopefully commit https://github.com/Smithay/client-toolkit/pull/25/commits/67076378f2d3d6dc86d0cefcf85175adfa53b6ea is more inline with what you were thinking. 

It tracks both pools buffers in a [Vec<Proxy<wl_buffer::WlBuffer>>; 2] with a pointer to the currently used pool (current_pool). When the draw is done it checks if the other pool is free (when the other pool's buffer vector length is 0) and if it is then it swaps the pools; if it cannot swap because the other pool is in use, then it sets pending_swap to true and the buffers will swap the pools when they are released.

vberger:
> 'Inactive' buffers are buffers that are contained in the pool thats not in use.

How can there be buffers in a pool if it is not in use? If there are non-released buffers pointing to a pool, then this pool may be in the process of being read by the compositor, and we should not write to it, or it may cause tearing or other graphical glitches.

-------

Thinking more about this, I think right way to implement this would rather be to change the`MemPool` / `DoubleMempool` utilities to directly integrate the tracking.

My idea would be that a `MemPool` would have an internal counter of how many buffers have been created from it and not been released. A shared `Arc<Mutex<usize>>` that is incremented every time a buffer is created and decremented every time a buffer is released. Meaning `MemPool::buffer(..)` would insert an implementation for the created buffer, and thus return a `Proxy<WlBuffer>` rather than a `NewProxy<WlBuffer>`. It would be a breaking change, but I need to plan the update to wayland-rs-0.21 which will be a breaking change too anyway.

To work correctly, this relies on the assumption that all created buffers will be commited and sent to the server for drawing, but there is hardly any point on creating a buffer to not use it...

With that, `MemPool` would gain a `is_used()` method, that returns `true` is this buffer count is greater than 0. If a `MemPool` is "used", this means that writing to it can corrupt what the compositor is currently reading and cause graphical glitches (and thus should be documented as such).

Then, `DoubleMemPool` would change to have its `pool` method return an `Option<&mut MemPool>`: if the first pool is not used it returns it, otherwise if the second one is not used it returns it, and otherwise (if both are used) it returns `None`. If it returns `None`, that means that it is not possible to draw right now and the draw should be delayed to later. The `swap` method could be removed too.

This way, `DoubleMemPool` would really be a proper wrapper to handle double-buffering. Then, using this new API, we can implement the double-buffering for the decoration drawing, and it should hopefully simplify a lot the logic.

What do you think of that?

trimental:
Sounds good to me, I thought about that but decided against it because it was a breaking change but if you're happy to break then its all good :) 

Am I right in thinking that pool1 and pool2 are the same but only pool2 gets rendered cause its created after pool1? Does the server spend time rendering both or does it know that pool2 eclipses pool1?

trimental:
Hopefully commit https://github.com/Smithay/client-toolkit/pull/25/commits/577c5a90a4fabbaca87efbf7b5e4cc5ced3181d1 does what you described. It does look a lot cleaner then before. 

The only problem I can think of is that you cannot destroy the buffer before a release event has been sent, so no 
```
buffers.drain(..) { buffer.destroy() } 
```
as it messes up the buffer count for the pool. No obvious solutions come to mind right now but I'll have a think about it.

src/window/basic_frame.rs
```diff
@@ -395,17 +389,13 @@ impl Frame for BasicFrame {
...
-        }
 
         {
             // grab the current pool
-            let mut pools = self.pools.lock().unwrap();
-            let pool = pools.pool();
+            let pool = match self.pools.pool() {
+                Some(pool) => pool,
+                None => return,
+            };
```
vberger: Here, to avoid completely missing a draw, there should be an internal boolean flag of the `Frame` that tracks that a draw was skipped because both mempools are in use. This flag would be reset to `false` whenever a redraw is really done.

Using the previously described callback of `DoubleMemPool`, whenever this callback fires, we check the flag in `Frame`. If it is set to `true`, that means that a draw was missed because the memory pools were in use. In this case, the `Frame` triggers a `FrameRequest::Refresh`, to ask the client to trigger a new redraw of the frame, to compensate for the one previously missed.

trimental: I implemented the frame's DoubleMemPool so that it calls a Refresh event without an internal boolean. This can be done because DoubleMemPool only fires when previously a None was returned from .pool() (which means you have to skip a draw) and a pool is now free. Am I right in thinking that?

vberger: Setting up `DoubleMemPool` to only fire its event if a `None` was previously returned from `pool()` sounds good, as long as it's clearly documented :+1:

src/utils/mempool.rs
```diff
@@ -49,15 +80,30 @@ impl DoubleMemPool {
...
+///
+/// Mempool internally tracks the lifetime of all buffers created from it and to ensure that
+/// this buffer count is correct all buffers must be attached to a surface. Once a buffer is attached to
+/// a surface it must be immediately commited to that surface before another buffer is attached.
+///
+/// Mempool will also handle the destruction of buffers and as such the `destroy()` method should not
+/// be used on buffers created from Mempool.
+///
+/// Mempool requires an implementation that will be called when the pool becomes free, this
+/// happens when all the pools buffers are released by the server.
```
vberger: Please add something like "Overwriting the contents of the memory pool before it is completely freed may cause graphical glitches as it may corrupt the data while the compositor is reading it."

src/window/basic_frame.rs
```diff
@@ -233,24 +230,31 @@ impl Frame for BasicFrame {
...
+            implem: Mutex::new(implementation),
+            maximized: Mutex::new(false),
+        });
+        let my_inner = inner.clone();
+        let pools = DoubleMemPool::new(&shm, move |_, _| {
+            my_inner
+                .implem
+                .lock()
+                .unwrap()
+                .receive(FrameRequest::Refresh, 0);
```
vberger: Please add a comment explaining that we can just blindly send a `Refresh` because the `DoubleMemPool` callback only fires if `DoubleMemPool.pool()` had previously returned `None`, which means a draw was postponed.

examples/image_viewer.rs
```diff
@@ -177,23 +175,14 @@ fn main() {
...
-    let free_pools = Rc::new(Cell::new(2u32));
     // - are we currently in the process of being resized? (to draw the image or
     //   black content)
     let mut resizing = false;
-    // - a QueueToken, we'll need it to implement the buffers
-    let token = event_queue.get_token();
     // - the size of our contents
     let mut dimensions = image.dimensions();
 
+    let mut buffer = None;
```
vberger: Given `MemPool` now destroys buffers by itself, we don't need to keep this variable any more.

examples/image_viewer.rs
```diff
@@ -262,85 +251,22 @@ fn main() {
...
+                None => {}
             }
-        } else if need_refresh {
-            // If we don't need to redraw but only to refresh do it
-            window.refresh();
-            // We still need to commit our surface, as the decorations are drawn as
-            // subsurfaces of it.
+            // We commit our drawing surface, so that the server atomically applies
+            // all the changes we previously did.
             window.surface().commit();
```
vberger: You moved the `commit()` into the `redraw` function, so I think this line can be removed.

examples/image_viewer.rs
```diff
@@ -262,85 +251,22 @@ fn main() {
...
-            need_refresh = false;
+            // We don't need to redraw or refresh anymore =)
+            need_redraw = false;
```
vberger: This should only be set to `false` if we did actually draw (so if `pool()` returned `Some(_)`.

trimental:
https://github.com/Smithay/client-toolkit/pull/25/commits/d5deca1c358dc705239f72a65dfc4185e31bb95b should fixup your review points. Feel free to be nitpicky with documentation as it should be pretty important for a breaking change.

vberger:
This needs a rebase as I merged #26, but other than that it is good :+1:

trimental:
Rebased to master

PR #26, opened 2018-08-02
Shm file

trimental:
After inspection of the GTK wayland backend I'm convinced that the memory file used by wayland for pixel buffering should be an anonymous shared memory file and not a temp file, [relevant GTK code](https://gitlab.gnome.org/GNOME/gtk/blob/master/gdk/wayland/gdkdisplay-wayland.c#L1229). The code I wrote uses memfd by default which is available on new linux kernel versions and shm_open as a fallback for older kernels. I think this is the standard way to implement the shared file for wayland.

### DIFF
--- a/src/utils/mempool.rs
+++ b/src/utils/mempool.rs
@@ -1,8 +1,12 @@
+use nix;
+use nix::fcntl;
+use nix::sys::mman;
+use nix::sys::stat;
 use std::fs::File;
 use std::io;
-use std::os::unix::io::AsRawFd;
+use std::os::unix::io::FromRawFd;
 
-use tempfile::tempfile;
+use rand::prelude::*;
 
 use wayland_client::protocol::{wl_buffer, wl_shm, wl_shm_pool};
 use wayland_client::{NewProxy, Proxy};
@@ -58,15 +62,39 @@ pub struct MemPool {
 impl MemPool {
     /// Create a new memory pool associated with given shm
     pub fn new(shm: &Proxy<wl_shm::WlShm>) -> io::Result<MemPool> {
-        let tempfile = tempfile()?;
-        tempfile.set_len(128)?;
+        let mut rng = thread_rng();
+        let mem_fd = loop {
+            let mem_file_handle = format!(
+                "/smithay-client-toolkit-{}",
+                rng.gen_range(0, ::std::u32::MAX)
+            );
+            match mman::shm_open(
+                mem_file_handle.as_str(),
+                fcntl::OFlag::O_CREAT
+                    | fcntl::OFlag::O_EXCL
+                    | fcntl::OFlag::O_RDWR
+                    | fcntl::OFlag::O_CLOEXEC,
+                stat::Mode::S_IRUSR | stat::Mode::S_IWUSR,
+            ) {
+                Ok(fd) => {
+                    mman::shm_unlink(mem_file_handle.as_str()).unwrap();
+                    break fd;
+                }
+                Err(nix::Error::Sys(nix::errno::Errno::EEXIST)) => continue,
+                Err(err) => panic!(err),
+            }
+        };
+        let mem_file = unsafe { File::from_raw_fd(mem_fd) };
+
+        mem_file.set_len(128)?;
+
         let pool = shm
-            .create_pool(tempfile.as_raw_fd(), 128)
+            .create_pool(mem_fd, 128)
             .unwrap()
             .implement(|e, _| match e {});
 
         Ok(MemPool {
-            file: tempfile,
+            file: mem_file,
             len: 128,
             pool: pool,
         })
--- a/src/utils/mempool.rs
+++ b/src/utils/mempool.rs
@@ -1,8 +1,10 @@
 use nix;
+use nix::errno::Errno;
 use nix::fcntl;
 use nix::sys::memfd;
 use nix::sys::mman;
 use nix::sys::stat;
+use std::ffi::CStr;
 use std::fs::File;
 use std::io;
 use std::os::unix::io::FromRawFd;
@@ -69,7 +71,7 @@ impl MemPool {
             rng.gen_range(0, ::std::u32::MAX)
         );
         let mem_fd = match memfd::memfd_create(
-            ::std::ffi::CStr::from_bytes_with_nul(b"smithay-client-toolkit\0").unwrap(),
+            CStr::from_bytes_with_nul(b"smithay-client-toolkit\0").unwrap(),
             memfd::MemFdCreateFlag::MFD_CLOEXEC,
         ) {
             Ok(fd) => fd,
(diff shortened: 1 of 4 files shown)


trimental:
[Relevant Qt code](https://github.com/qt/qtbase/blob/a37dd93defd91b79fb6730d0ff0515a66a0d3972/src/corelib/kernel/qsharedmemory_posix.cpp#L95)

trimental:
Its worth noting that QtWayland does fallback onto temp files if it cannot create a shared memory file properly with memfd, however the standard seems to be the linux way of creating memory files.

vberger:
This seems like a clean handling of things, thanks!

However, could you factor the logic of "creating the tempfile" into its own function? Something like

```rust
fn create_shm_fd() -> io::Result<RawFd> { ... }
```

and don't panic if it fails to create the tempfile, but rather return the error.

trimental:
Yeah I'll move the code into a function. However with the returning of the error, is it possible to return a nix::Error to a io::Result? Also do you want the function to be apart of the MemPool impl or separate?

vberger:
There is an `impl From<nix::errno::Errno> for std::io::Error`, which allows you to handle the the `Sys` variant of `nix::Error`. The other variants should not happen, and you can panic for those.

vberger:
And please make it a standalone function. :+1:

trimental:
I'll warn that the code is a kind of a web of matching statements, this can be less 'indented' and 'matchy' if you use a bool to track whether to fallback or not. I think its a bit of a preference thing and I left it 'matchy' to avoid allocating a bool but if you want I can rewrite it like that so its more readable.

src/utils/mempool.rs
```diff
@@ -182,3 +145,52 @@ impl io::Seek for MemPool {
...
+            Ok(fd) => break Ok(fd),
+            Err(nix::Error::Sys(Errno::EINTR)) => continue,
+            Err(nix::Error::Sys(errno)) => break Err(io::Error::from(errno)),
+            Err(_) => {
+                // Fallback to using shm_open
+                let mut mem_file_handle = format!(
+                    "/smithay-client-toolkit-{}",
+                    rng.gen_range(0, ::std::u32::MAX)
+                );
+                break loop {
```
vberger: I think this block can be flattened like so:

```rust
loop {
    match memfd_create(...) {
        Ok(fd) => return Ok(fd),
        Err(nix::Error::Sys(Errno::EINTR)) => continue,
        Err(nix::Error::Sys(errno)) => return Err(io::Error::from(errno)),
        Err(_) => break,
    }
}
loop {
    /* fallback to using shm_open */
}
```

trimental: Nice catch

src/utils/mempool.rs
```diff
@@ -137,3 +145,51 @@ impl io::Seek for MemPool {
...
+            mem_file_handle.as_str(),
+            fcntl::OFlag::O_CREAT
+                | fcntl::OFlag::O_EXCL
+                | fcntl::OFlag::O_RDWR
+                | fcntl::OFlag::O_CLOEXEC,
+            stat::Mode::S_IRUSR | stat::Mode::S_IWUSR,
+        ) {
+            Ok(fd) => match mman::shm_unlink(mem_file_handle.as_str()) {
+                Ok(_) => return Ok(fd),
+                Err(nix::Error::Sys(errno)) => return Err(io::Error::from(errno)),
```
vberger: Here we need to close the fd, or we'll leak it.

trimental: Like with [unistd::close](https://docs.rs/nix/0.11.0/nix/unistd/fn.close.html)? Does shm_unlink not do this?

vberger: yes, using `close()`. `shm_unlink` removes the link of the SHM to the global namespace, preventing other processes from accessing it, but does not close the file descriptor, according to the man page:

> The operation of shm_unlink() is analogous to unlink(2): it removes a shared memory object name, and, once all processes have unmapped  the  object, de-allocates  and destroys the contents of the associated memory region.

trimental: Closing the fd leads to a bad file descriptor error, maybe its the close or maybe its the wrapping of the fd into a file which closes the fd on drop.

vberger: Where is it wrapped into a `File`? I just meant closing it in the `Err(nix::Error::Sys(errno))` branch of the `shm_unlink` match, and this branch only, as in this case we do not return the FD from the function...

trimental: Right, I got a bit confused and thought you were talking about the Ok(_) branch

vberger:
Could you please not insert `cargo fmt` commits changing the whole project in the PR? It makes the diff history hard to read. I'll make a fmt-dedicated commit.

trimental:
Sorry, I'll remove that commit

src/utils/mempool.rs
```diff
@@ -138,3 +145,54 @@ impl io::Seek for MemPool {
...
+
+fn create_shm_fd() -> io::Result<RawFd> {
+    loop {
+        match memfd::memfd_create(
+            CStr::from_bytes_with_nul(b"smithay-client-toolkit\0").unwrap(),
+            memfd::MemFdCreateFlag::MFD_CLOEXEC,
+        ) {
+            Ok(fd) => return Ok(fd),
+            Err(nix::Error::Sys(Errno::EINTR)) => continue,
+            Err(nix::Error::Sys(errno)) => return Err(io::Error::from(errno)),
```
vberger: Actually, I think we should break on `Err(nix::Error::Sys(Errno::ENOSYS))` (which means that `memfd_create` is not supported by the current kernel, and thus we should fallback on shm_open, if I believe the [GTK code](https://gitlab.gnome.org/GNOME/gtk/blob/master/gdk/wayland/gdkdisplay-wayland.c#L1212-1216)), and the final `Err(_)` catch-all should likely be marked `unreachable!()`, as no nix-specific errors can be generated by this function.

src/utils/mempool.rs
```diff
@@ -138,3 +145,54 @@ impl io::Seek for MemPool {
...
+                // If a file with that handle exists then change the handle
+                mem_file_handle = format!(
+                    "/smithay-client-toolkit-{}",
+                    rng.gen_range(0, ::std::u32::MAX)
+                );
+                continue;
+            }
+            Err(nix::Error::Sys(Errno::EINTR)) => continue,
+            Err(nix::Error::Sys(errno)) => return Err(io::Error::from(errno)),
+            Err(err) => panic!(err),
```
vberger: same here, please use `unreachable!(err)`

src/utils/mempool.rs
trimental: Nix::Error can return errors other then the Sys enum which means Err(_) isn't really unreachable

trimental: I've used panic!(err) instead of unreachable!(err) because its not unreachable and it does the same thing as panic.

src/utils/mempool.rs
vberger: This branch (and the similar one on the other match) covers nix-specific errors that supposedly cannot occur in our case. As such I'd prefer to use the `unreachable!(err)` macro instead of a regular `panic!(err)`, to better convey this.

trimental: https://github.com/Smithay/client-toolkit/pull/26/commits/978b03742abdaaac18512fa3218efb9be51effbb should do this :)

vberger:
Alright, this is good, thanks! :+1:

PR #27, opened 2018-08-04
Fix travis ci rustfmt check

trimental:
This PR fixes the travis rustfmt check.


vberger:
Ah right, `--write-mode=diff` non longer exits with failure if there is a diff... Thanks.

PR #28, opened 2018-08-06
Fix window resizing

trimental:
### Issue
Resizing from the left side of the window causes incorrect behavior
Introduced in https://github.com/Smithay/client-toolkit/commit/84bb8e22323bccf5c181a0f08e5a97f6db342c9e
 
### Fix
Set minimum window size to 2px

I'll be honest in saying that I don't really know why this fixes the issue (especially since the top, bottom and right side works) but it seems to fix it perfectly, at least on mutter as it may be compositor specific.


vberger:
What is the incorrect behavior that was introduced exactly?

trimental:
When you move the left handle past the right side of the window, the window will glitch position towards the right (move right)

trimental:
This also seems to fix an issue I had with alacritty panicking at random times when resizing (even without moving the left handle). It returned
```
wl_display@1: error 1: invalid arguments for zwp_linux_buffer_params_v1@271.create_immed
thread 'pty reader' panicked at 'called `Result::unwrap()` on an `Err` value: EventsLoopClosed', libcore/result.rs:945:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
thread 'main' panicked at 'Wayland connection lost.: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }', libcore/result.rs:945:5
```

trimental:
It looks to me like this could be a bug in the mutter compositor where 1px width windows aren't supported (I may be wrong). Could you please test whether the issue I described happens in master on any other compositor, if it does occur then some code in the client-toolkit is probably wrong but if it doesn't then this will indicate its a bug in mutter. If it is a bug in mutter you can choose whether you want to set 2px as the minimum width (doubt it will disrupt anything) or wait until the bug is fixed in mutter (and close this).

vberger:
I cannot reproduce the bug you describe on neither weston nor wlroots using the master branch, so I suspect this is likely indeed a mutter bug, so it is worth reporting to them in all cases.

On the other hand, setting a minimum width to 2 is a minor mitigation that is unlikely to affect anything wrongly.

But if we start introducing a nontrivial minimal size, I'd rather it be a common constant `MIN_WINDOW_SIZE` that is reused everywhere, with a comment explaining why it has this value, ideally with a like to the relevant mutter bug.

trimental:
I've filed a [issue with mutter](https://gitlab.gnome.org/GNOME/mutter/issues/259), moved minimum size into a const, updated the changelog and rebased to master. 

Even if this issue gets sorted quickly it might be best to wait a little while after the next stable release of mutter that the fix rolls out in so that it can be assumed most people will have updated, this is probably when the major distros update.

src/window/mod.rs
```diff
@@ -22,6 +22,9 @@ mod shell;
 
 pub use self::basic_frame::BasicFrame;
 
+// Defines the minimum window size. Minimum width is set to 2 pixels to circumvent a bug in mutter - https://gitlab.gnome.org/GNOME/mutter/issues/259
```
vberger: Could you split this comment into two lines? It's very long with the URL

PR #29, opened 2018-08-10
Fix reasonable clippy lints

trimental:
This PR fixes some of the 'reasonable' clippy issues. It fixes quite a few clippy lints however I left a few out because they were either irrelevant or unreasonable, for example passing a reference to a proxy and shortening arguments to functions, etc.

Heres a list of general clippy issues it fixes
* use `if lets` over `match`
* removal of \`static from consts (which are already static) - `&'static [u8] -> &[u8]`
* shorten argument passing - `some_arg: some_arg -> some_arg`
* removal of redundant returns - `return Err() -> Err()`
* removal of `clone()` for objects with Copy
* check if result `is_ok()` instead of matching - `match result -> if result.is_ok()`
* space out numbers - `1000000 -> 1_000_000`
* use AtomicBool over Arc<Mutex<bool>>
* pass reference instead of object
* use `from()` instead of `as`
* use `if else` instead of if `{ else {} }`

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -107,16 +107,14 @@ impl Environment {
                     id,
                     ref interface,
                     version,
-                } => match &interface[..] {
-                    "wl_output" => outputs2.new_output(
+                } => if let "wl_output" = &interface[..] {
+                    outputs2.new_output(
                         id,
                         registry.bind::<wl_output::WlOutput>(version, id).unwrap(),
-                    ),
-                    _ => (),
+                    )
                 },
-                GlobalEvent::Removed { id, ref interface } => match &interface[..] {
-                    "wl_output" => outputs2.output_removed(id),
-                    _ => (),
+                GlobalEvent::Removed { id, ref interface } => if let "wl_output" = &interface[..] {
+                    outputs2.output_removed(id)
                 },
             }
             cb.receive(event, registry);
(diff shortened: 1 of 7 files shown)


vberger:
Looks good overall, I'm just wary about using `AtomicBool` over `Mutex<bool>`. The semantics are not the same: if I lock a mutex and access its contents 2 times, I am guaranteed that the contents won't change between the two accesses. This is not true with an atomic, and I suspect the code might rely on this.

So, I'd rather merge this without the atomic/mutex change, unless you can explain to me that the change does not affect the behavior.

trimental:
I just assumed that AtomicBool was the same as Mutex but less heavy because clippy suggest it. 

I found that DoubleMemPool's `free` should be a mutex otherwise the DoubleMemPool implementation might get called twice and also that `resizable.load()` shouldn't be stored in a variable but rather dynamically called. Other then that I think it's safe.

1. `resizable.load()` is only called when checking whether to call `change_pointer()` and when its value is passed to `request_for_location()`. Neither of these two behaviors would be different if the AtomicBool was changed during the operation.

2. `maximized.load()` is used when it is set, in which case even if the value is changed it will be overwritten; when it is used to check whether to draw the rounded header edges; and when it is passed to `request_for_location()`

that said, I only implemented AtomicBool instead of Mutex because clippy suggested it, if you think it would be easier to maintain and understand code with Mutexs then I think that's fine as well because it makes my brain spin a little to think about which situations non-locking is acceptable. I'll wait for your answer on AtomicBools vs Mutexs before fixing the first two issues I mentioned.

vberger:
I guess the simplest is to stick with mutexes. I doubt the performance impact would be measurable anyway.

PR #32, opened 2018-08-22
End key repetition when keyboard loses focus

trimental:
This PR ends key repetition when the keyboard loses focus as the release event will never be sent in these cases. When the keyboard loses focus it just ends the key repetition and doesn't send a release event as the key isn't actually released (in the physical sense) I can change this behavior if you want.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


vberger:
This is good like that, we should not introduce a `release` event given we already forward the `leave` event.

Could you just add a changelog entry with that?

PR #33, opened 2018-09-13
Upgrade to wayland-rs 0.21

trimental:
This PR upgrades the smithay-client-toolkit to use wayland-rs 0.21 (pure rust implementation :tada:)

This PR makes necessary breaking modifications to the API but I've tried to minimize the changes, the API can be changed to better fit wayland-rs 0.21 if you'd like. I've run this through clippy and tested the examples but I'll be porting winit to use this PR to test the changes in more depth as well as reading through the diff a bit more.

### DIFF
--- a/src/data_device/data_device.rs
+++ b/src/data_device/data_device.rs
@@ -111,16 +111,12 @@ pub enum DndEvent<'a> {
     },
 }
 
-fn data_device_implem<Impl>(
-    event: wl_data_device::Event,
-    inner: &Arc<Mutex<Inner>>,
-    implem: &mut Impl,
-) where
-    for<'a> Impl: FnMut((), DndEvent<'a>),
+fn data_device_implem<Impl>(event: wl_data_device::Event, inner: &mut Inner, implem: &mut Impl)
+where
+    for<'a> Impl: FnMut(DndEvent<'a>),
 {
     use self::wl_data_device::Event;
 
-    let mut inner = inner.lock().unwrap();
     match event {
         Event::DataOffer { id } => inner.new_offer(id),
         Event::Enter {
@@ -131,36 +127,27 @@ fn data_device_implem<Impl>(
             id,
         } => {
             inner.set_dnd(id);
-            implem(
-                (),
-                DndEvent::Enter {
-                    serial,
-                    surface,
-                    x,
-                    y,
-                    offer: inner.current_dnd.as_ref(),
-                },
-            );
+            implem(DndEvent::Enter {
+                serial,
+                surface,
+                x,
+                y,
+                offer: inner.current_dnd.as_ref(),
+            });
         }
         Event::Motion { time, x, y } => {
-            implem(
-                (),
-                DndEvent::Motion {
-                    x,
-                    y,
-                    time,
-                    offer: inner.current_dnd.as_ref(),
-                },
-            );
+            implem(DndEvent::Motion {
+                x,
+                y,
+                time,
+                offer: inner.current_dnd.as_ref(),
+            });
         }
-        Event::Leave => implem((), DndEvent::Leave),
+        Event::Leave => implem(DndEvent::Leave),
         Event::Drop => {
-            implem(
-                (),
-                DndEvent::Drop {
-                    offer: inner.current_dnd.as_ref(),
-                },
-            );
+            implem(DndEvent::Drop {
+                offer: inner.current_dnd.as_ref(),
+            });
         }
         Event::Selection { id } => inner.set_selection(id),
     }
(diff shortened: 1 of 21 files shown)


src/pointer/mod.rs
```diff
@@ -55,25 +56,33 @@ impl AutoThemer {
...
+    pub fn theme_pointer_with_impl<Impl, UD>(
         &self,
-        pointer: NewProxy<wl_pointer::WlPointer>,
+        seat: &Proxy<wl_seat::WlSeat>,
         mut implementation: Impl,
+        user_data: UD,
     ) -> AutoPointer
     where
-        Impl: Implementation<AutoPointer, wl_pointer::Event> + Send + 'static,
+        Impl: FnMut(AutoPointer, wl_pointer::Event) + Send + 'static,
```
vberger: Same here for the order of arguments.

src/pointer/theme.rs
```diff
@@ -32,7 +32,7 @@ impl ThemeManager {
     pub fn init(
         name: Option<&str>,
         compositor: Proxy<wl_compositor::WlCompositor>,
-        shm: Proxy<wl_shm::WlShm>,
+        shm: &Proxy<wl_shm::WlShm>,
```
vberger: Why this change?

src/utils/mempool.rs
```diff
@@ -41,7 +40,7 @@ impl DoubleMemPool {
     /// Create a double memory pool
     pub fn new<Impl>(shm: &Proxy<wl_shm::WlShm>, implementation: Impl) -> io::Result<DoubleMemPool>
     where
-        Impl: Implementation<(), ()> + Send,
+        Impl: FnMut((), ()) + Send + 'static,
```
vberger: The `()` arguments can be removed here too.

src/window/basic_frame.rs
```diff
@@ -228,7 +229,7 @@ impl Frame for BasicFrame {
         compositor: &Proxy<wl_compositor::WlCompositor>,
         subcompositor: &Proxy<wl_subcompositor::WlSubcompositor>,
         shm: &Proxy<wl_shm::WlShm>,
-        implementation: Box<Implementation<u32, FrameRequest> + Send>,
+        implementation: Box<FnMut(u32, FrameRequest) + Send>,
```
vberger: I'd prefer the arguments of this closure swapped to. Main matter (the requests) first, additional data second.

src/window/mod.rs
```diff
@@ -60,7 +59,7 @@ pub enum Event {
 struct WindowInner<F> {
     frame: Arc<Mutex<F>>,
     shell_surface: Arc<Box<shell::ShellSurface>>,
-    user_impl: Box<Implementation<(), Event> + Send>,
+    user_impl: Box<FnMut((), Event) + Send>,
```
vberger: The `()` arugment can be dropped here too.

src/data_device/data_device.rs
```diff
@@ -267,7 +275,7 @@ impl DataDevice {
     ///
     /// Correspond to traditional copy/paste behavior. Setting the
     /// source to `None` will clear the selection.
-    pub fn set_selection(&self, source: Option<DataSource>, serial: u32) {
+    pub fn set_selection(&self, source: &Option<DataSource>, serial: u32) {
```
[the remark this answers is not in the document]
vberger: I'd prefer to keep this PR focused on the upgrade, and leave the clippy improvements as an other PR, to handle these questions separately.

trimental:
I'll go through the documentation where changes were made

src/env.rs
```diff
@@ -74,50 +71,44 @@ impl Environment {
...
     ) -> io::Result<Environment>
     where
-        Impl: Implementation<Proxy<wl_registry::WlRegistry>, GlobalEvent> + Send + 'static,
+        Impl: FnMut(GlobalEvent, Proxy<wl_registry::WlRegistry>) + Send + 'static,
     {
         let outputs = ::output::OutputMgr::new();
         let outputs2 = outputs.clone();
 
-        let manager = GlobalManager::new_with_cb(registry, move |event, registry: Proxy<_>| {
+        let manager = GlobalManager::new_with_cb(&display, move |event, registry| {
```
vberger: Actually, for this to be correct now that we create the registry ourselves, we need to ensure that the registry will be created on the event queue that was given as argument.

To ensure that, we need to first make a wrapper of the display:

```rust
let display_wrapper = display.make_wrapper(&evq.get_token());
```

and then give this wrapper to `GlobalManager::new_with_cb(..)`

vberger:
Okay, there this one last thing, and with that I think it'll be good. Thanks! :+1:

trimental:
Looks like rust fmt has changed in 1.29 but I'll leave the formatting for a separate 'Cargo fmt' commit

PR #34, opened 2018-09-16
Add `keyboard::Event::Modifiers`

trimental:
When a modifier key is pressed then the modifiers will not show that key as active, for example pressing the shift key will show shift as inactive in the `Event::Key { modifiers, .. }`, therefore a modifiers event allows access to modifiers even when that modifier is the key being pressed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


vberger:
I was hoping we could get away without this event, but it looks like that is wrong.

If we add it, I guess we can remove the `modifiers` field of the other events of the enum though, given they'd mostly be redundant.

PR #35, opened 2018-09-17
Display title in the window header decoration

trimental:
This is a work in progress PR for titled windows. It currently uses rusttype and fontconfig-rs to draw text and select fonts.

screenshot from 2018-09-17 18-25-46

This PR currently has some issues with the loading of '.otf' fonts (this seems to be to do with rusttype) as well as the use of font_loader which depends on fontconfig-rs meaning projects using both the client-toolkit and fontconfig-rs will get a cargo multiple native link error.

Are titled windows something you think would be nice to have BasicFrame support or do you think its too complex?

- [x] Better selection of window fonts (using gsettings and kde equivalent to find default window title font) (wont fix for the time being)
- [ ] Have the title shrink to 'part_of_title...' (for example 'alacrit...') instead of completely disappearing
- [x] Try to remove fontconfig-rs dependency
- [x] Investigate rusttype '.otf' support (going to just use '.ttf' for now)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- BasicFrame: Display the title of the window in the window header
 - Pass `set_selection()` `Option<DataSource>` and `AutoThemer::init()` `Proxy<WlShm>` by reference
 - Window: add `set_theme()` function which takes an object implementing the trait `Theme` to adjust the look of window decorations
 - Window: add new `ConceptFrame` which provides an alternative to the `BasicFrame` window decorations
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ dlib = "0.4"
 lazy_static = "1"
 memmap = "0.6"
 rand = "0.5"
-andrew = "0.1.1"
+andrew = "0.1.2"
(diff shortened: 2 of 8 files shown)


vberger:
Looks pretty good on the principle :+1: 

I agree this would need a better selection of fonts, or at least a reasonable fallback mechanism to not depend on any particular font.

> Try to remove fontconfig-rs dependency

Could you explain why?

trimental:
> at least a reasonable fallback mechanism to not depend on any particular font

Actually 'fontconfig match' will always return a font (the default font if it can't find the font you gave it), I just have the specific fonts to try and match the right size and stuff, we can fallback just on whatever 'fontconfig match' gives us

> Try to remove fontconfig-rs dependency

I don't really have much experience with this cargo error but when I try to run alacritty with this PR cargo gives me 
```
error: multiple packages link to native library `fontconfig`, but a native library can be linked only once
```

vberger:
Hmm, looks like alacritty uses its own fork of libfontconfig: https://github.com/jwilm/alacritty/blob/master/font/Cargo.toml#L15

This means it cannot interact with any other crate using fontconfig, unless they depend themselves on the same fork...

PR #38, opened 2018-10-02
Cleanup - Handle function errors, remove unnecessary assignment, pass arguments more efficiently

trimental:
I think its better to panic on functions that return results even if it doesn't matter too much if the function fails, or at least they should be printed to stderr to log the failure.

A few arguments can be passed more efficiently, proxies by reference probably don't save a lot performance wise but are more idiomatic and now's probably a good time to change these considering the next release will be breaking.

### DIFF
--- a/src/window/basic_frame.rs
+++ b/src/window/basic_frame.rs
@@ -521,7 +521,12 @@ impl Frame for BasicFrame {
                         *b = 0x00;
                     }
                 }
-                let _ = mmap.flush();
+                if let Err(err) = mmap.flush() {
+                    eprintln!(
+                        "[SCTK] Basic frame: failed to flush frame memory map: {}",
+                        err
+                    );
+                }
             }
 
             // Create the buffers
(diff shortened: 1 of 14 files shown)


vberger:
Sorry for the delay on coming back on this.

For errors that are non-fatal (for example failing to draw, it *could* be an intermittent error ?), I'd rather just print something on stderr and continue, rather than panic.

src/window/basic_frame.rs
```diff
@@ -288,14 +288,12 @@ impl Frame for BasicFrame {
...
-                            change_pointer(&pointer, data.location, Some(serial));
+                            change_pointer(&pointer, data.location, Some(serial))
                         }
                     }
                     Event::Leave { serial, .. } => {
                         data.location = Location::None;
-                        if resizable {
-                            change_pointer(&pointer, data.location, Some(serial));
-                        }
+                        change_pointer(&pointer, data.location, Some(serial))
```
vberger: Why was this `if` removed ?

trimental: Ah, to be honest I think this is just a mistake I didn't catch in the git diff

vberger:
Regarding printing messages to stderr, I tend to prefix the messages according to the crate generating them, and add some information about where the error originated. So something like:

`[SCTK] Data offer: failed to close write pipe: {:?}`

PR #45, opened 2018-11-21
WIP: Add surface abstraction.

### LINKED ISSUE #44: Scope of client-toolkit
What is the scope of client-toolkit? It seems to me that some of the stuff that is handled in winit could be moved over here, like utilities for hidpi handling for example.

I could just use winit of course but I'm interested in experimenting with a new wayland protocol for creating menus, like for example in macosx where they are part of the window manager. In particular I want to create a spacemacs style compositor.

dvc94ch:
Still need to do some more testing

Closes #44

### DIFF
--- a/src/surface.rs
+++ b/src/surface.rs
@@ -1,0 +1,85 @@
+use std::sync::{Arc, Mutex};
+use wayland_client::Proxy;
+use wayland_client::protocol::{wl_output, wl_surface};
+use wayland_client::protocol::wl_compositor::{RequestsTrait as CompositorRequest};
+use crate::env::Environment;
+use crate::output::OutputMgr;
+
+pub struct Surface {
+    surface: Proxy<wl_surface::WlSurface>,
+    inner: Arc<Mutex<InnerSurface>>
+}
+
+struct InnerSurface {
+    output_manager: OutputMgr,
+    outputs: Vec<Proxy<wl_output::WlOutput>>,
+    scale_factor: i32,
+    new_scale_factor: Option<i32>,
+}
+
+impl InnerSurface {
+    pub fn new(env: &Environment) -> Self {
+        InnerSurface {
+            output_manager: env.outputs.clone(),
+            outputs: Vec::new(),
+            scale_factor: 1,
+            new_scale_factor: None,
+        }
+    }
+
+    pub fn scale_factor(&self) -> i32 {
+        self.scale_factor
+    }
+
+    pub fn set_scale_factor(&mut self, scale_factor: i32) {
+        if self.scale_factor != scale_factor {
+            self.scale_factor = scale_factor;
+            self.new_scale_factor = Some(scale_factor);
+        }
+    }
+}
+
+impl Surface {
+    pub fn new(env: &Environment) -> Self {
+        let inner = Arc::new(Mutex::new(InnerSurface::new(env)));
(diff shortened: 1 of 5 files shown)


dvc94ch:
Updated the PR. I think it's much better now and also nicer than the winit implementation.

src/surface.rs
```diff
@@ -0,0 +1,53 @@
+//! Utility functions for creating dpi aware wayland surfaces.
+
+use std::sync::{Arc, Mutex};
+use wayland_client::Proxy;
+use wayland_client::protocol::wl_surface;
+use wayland_client::protocol::wl_compositor::{RequestsTrait as CompositorRequest};
+use crate::env::Environment;
+
+/// Creates a WlSurface from an Environment.
+/// Takes a callback for notification of dpi changes.
```
vberger: Please add a newline between the title of the description and its body. Also please expand the documentation to detail how the dpi value is computed (taking the max of the dpi factors of all outputs this surface is displayed on).

src/surface.rs
```diff
@@ -0,0 +1,53 @@
...
+pub fn create_surface<F, T>(
+    environment: &Environment,
+    dpi_change: F,
+) -> Proxy<wl_surface::WlSurface>
+where
+    F: Fn(i32) -> T + Send + 'static
```
vberger: I think this could be an `FnMut` rather than a plain `Fn`, couldn't it?

src/surface.rs
```diff
@@ -0,0 +1,53 @@
...
+                }
+                dpi_change(scale_factor);
+            }
+        }, Mutex::new(1))
+    }).unwrap()
+}
+
+/// Returns the current dpi factor of a surface.
+pub fn get_dpi_factor(surface: &Proxy<wl_surface::WlSurface>) -> i32 {
+    *surface.user_data::<Mutex<i32>>().unwrap().lock().unwrap()
```
vberger: We could handle more gracefully the case where the user tries to use this method on a surface that was not created by this helper, in which case the `user_data()` call will return `None`. A possibility would be to replace the first `unwrap()` with an `expect(..)` with a descriptive error message.

src/surface.rs
```diff
@@ -0,0 +1,53 @@
+//! Utility functions for creating dpi aware wayland surfaces.
+
+use std::sync::{Arc, Mutex};
+use wayland_client::Proxy;
+use wayland_client::protocol::wl_surface;
+use wayland_client::protocol::wl_compositor::{RequestsTrait as CompositorRequest};
+use crate::env::Environment;
```
vberger: SCTK tries to remain compatible with older versions of rust, as such we can't use features like `crate::`

vberger:
Perfect! I think this is just missing an entry in the changelog before merging.

vberger:
Looks like there are still some small issues, the travis build is still legitimately red

vberger:
Perfect! Thanks a lot for taking the time to implement this! :+1:

vberger:
With some refactoring (putting the outputs list in the user data) it would be pretty easy to expose a freestanding function to retrieve the list of monitors. Something like

```
fn get_outputs(&Proxy<WlSurface>) -> Vec<Proxy<WlOutput>> { ... }
```

And it' be reasonable to change the closure to pass the surface as argument too.

PR #46, opened 2018-11-22
Add output list to surface and pass surface to dpi changed closure.

dvc94ch:


src/surface.rs
```diff
@@ -44,23 +62,40 @@ where
...
         .expect("Surface was not created with create_surface.")
+        .dpi_factor
         .lock()
         .unwrap()
 }
+
+/// Returns a list of outputs the surface is displayed on.
+pub fn get_outputs<'a>(
+    surface: &'a Proxy<wl_surface::WlSurface>,
+) -> RwLockReadGuard<Vec<Proxy<wl_output::WlOutput>>> {
```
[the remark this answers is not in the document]
dvc94ch: Probably because I was wondering how I'd return a readonly vec from a mutex. I think the answer is to not use a mutex =P But cloning works too

src/surface.rs
```diff
@@ -16,51 +30,65 @@ pub fn create_surface<F>(
...
+    surface
+        .user_data::<Mutex<SurfaceUserData>>()
+        .expect("Surface was not created with create_surface.")
+        .lock()
+        .unwrap()
+        .dpi_factor
+}
+
+/// Returns a list of outputs the surface is displayed on.
+pub fn get_outputs<'a>(
```
vberger: You don't need this lifetime annotation any longer =)

vberger:
Ah yes right it was what https://github.com/tomaka/winit/pull/719 is about.

This is going to be difficult to manage though. Compositors should be sending the `Leave` events in my opinion.

dvc94ch:
We could make create_surface a method of Environment and track the surfaces. Since it requires an Environment anyway we are relying on that abstraction for the create_surface abstraction anyway. 

I guess filing a bug report in gnome won't do much good since we'll have to support this hack for the forseeable future anyway for all broken gnome versions...

dvc94ch:
Is there a way to dispatch an event in a Proxy? Then we can fake the leave to all surfaces when an output is destroyed.

vberger:
No, this is not possible. You'd need to keep track of all the surfaces, and access their user data to update their monitor list and dpi factor whenever a monitor is removed. This in turn means you'll need to store the callback in the user data as well to be able to call it.

src/env.rs
```diff
@@ -195,11 +216,22 @@ impl Environment {
...
         })
     }
 
     /// Retrieve the accepted SHM formats of the server
     pub fn shm_formats(&self) -> Vec<wl_shm::Format> {
         self.shm_formats.lock().unwrap().clone()
     }
+
+    /// Create a new dpi aware surface
+    pub fn create_surface<F>(&self, dpi_change: Box<F>) -> Proxy<wl_surface::WlSurface>
```
vberger: Please do the boxing of the callback internally, and let the public method take non-boxed closure.

vberger:
Looks good to me, with the small nit. :+1:

Issue #47, opened 2018-11-23
Support loading dpi-scaled cursor images


vberger:
The `ThemedPointer` should be improved to support dpi-scaled icons loading.

kchibisov:
Do we want to add something like 
```rust
pub fn theme_pointer_with_dpi(&self, pointer: Proxy<wl_pointer::WlPointer>, dpi: u32) -> ThemedPointer
``` 
and new constructor for ThemeManager or to add just one dpi aware constructor for ThemeManager and create new pointers with this dpi auto applied?

vberger:
I think the simplest & most consistent way would be to just have cursors always be dpi-aware, using the dpi-tracking facility already in SCTK.

Though, this is mostly adding more inter-coupling between the modules of SCTK. At this point I think this is illusory to maintain the very-loose coupling that is the current state...

kchibisov:
Hm, just looking through the window code(basic_frame and concept_frame) it seems like they use AutoThemer, what's the use case of AutoThemer etc without Window(surface)?

vberger:
AutoThemer provides the necessary tools to change the cursor image. The window code uses it to change the cursor for resizing interactions.

One thing to be aware of is that in Wayland, a surface and a window are two quite different things. A surface is just a canvas the client draws on, and for it to be displayed it must be assigned a role. The two most common roles are "toplevel" (aka: window), and "pointer", a cursor image.

So an app build using SCTK will actually manipulate several surfaces: the surface for its main content, the surface manager by the `AutoThemer` for cursor icons, and a few other surfaces used to draw the decorations by ConceptFrame.

Now, writing all that I realize that we actually cannot have the Themer track the DPI by itself, because it does not have access to a surface to track it. So I see two approaches:

- setup the Themer to monitor the list of outputs (using an handle to the `OutputMgr`), compute the highest dpi factor of all outputs, and always use this one. Meaning the compositor will have to downsample the cursor icon on the lower-dpi screens. Which would likely have a rather low cost in practice given they are small surfaces.
- add a `set_cursor_with_dpi(...)` method to `ThemedPointer`/`AutoPointer` allowing the user to specify the required dpi when changing the pointer. The user is now responsible for knowing which DPI factor it wants the cursor to be user with.

In the second case, `ThemeManager` would thus load the cursors themes lazily for any required dpi factor, and cache the dpi-instanciated themes for later use, it'd say.

kchibisov:
[...]

PR #48, opened 2018-11-23
Pointer dpi example

dvc94ch:
Updates the pointer_input example. Relates to #47 

So figured out what needs to happen for #47. It looks like we can set_buffer_scale(max_dpi_of_all_outputs) and load the cursor theme when we get a GlobalEvent::New or GlobalEvent::Removed and let the compositor downscale the cursor when on a lowdpi screen.

### DIFF
--- a/examples/pointer_input.rs
+++ b/examples/pointer_input.rs
@@ -7,7 +7,6 @@ use std::sync::{Arc, Mutex};
 
 use byteorder::{NativeEndian, WriteBytesExt};
 
-use sctk::reexports::client::protocol::wl_compositor::RequestsTrait as CompositorRequests;
 use sctk::reexports::client::protocol::wl_seat::RequestsTrait as SeatRequests;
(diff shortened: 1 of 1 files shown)


dvc94ch:
I think the simplest solution is to have a buffer_scale of 4 by default for all cursors and then let the compositor scale it. What do you think?

vberger:
Please refrain from packing several changes unrelated to each other into the same pull request, this makes it quite difficult to review. If this PR is about updating the pointer-input example, please leave the other changes for an other PR.

Also, please do not change the public APIs of the crate if it is not necessary.

dvc94ch:
Well this is essentially about fixing #47 and updating the example. Other than using an oversized buffer_scale and letting the compositor downscale the pointer, I'm not sure how not to change the public api. It seems we need at least an output manager if not an environment to handle cursor dpi scaling properly.

PR #49, opened 2018-11-24
Dpi cursor scaling.

### LINKED ISSUE #47: Support loading dpi-scaled cursor images

dvc94ch:
Fixes #47 without changeing the public api for a dpi factor <= 4.

### DIFF
--- a/src/pointer/theme.rs
+++ b/src/pointer/theme.rs
@@ -40,7 +40,7 @@ impl ThemeManager {
 
         Ok(ThemeManager {
             compositor,
-            theme: Arc::new(Mutex::new(load_theme(name, 16, &shm))),
+            theme: Arc::new(Mutex::new(load_theme(name, 16 * 16, &shm))),
         })
     }
 
@@ -50,6 +50,8 @@ impl ThemeManager {
             .compositor
             .create_surface(|surface| surface.implement(|_, _| {}, ()))
             .unwrap();
+        surface.set_buffer_scale(4);
+
         ThemedPointer {
             pointer,
             inner: Arc::new(Mutex::new(PointerInner {
@@ -79,6 +81,7 @@ impl ThemeManager {
             .compositor
             .create_surface(|surface| surface.implement(|_, _| {}, ()))
             .unwrap();
+        surface.set_buffer_scale(4);
 
         let inner = Arc::new(Mutex::new(PointerInner {
             surface,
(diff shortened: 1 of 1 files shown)


vberger:
Ok. So, if I understand correctly the reason you needed to change the API was in order to access the output manager from the `Environment` to track the DPI of the outputs, right?

I'd like to keep changes backward-compatible as much as possible for now, and many parts of the crate can still be used even if the user of SCTK decides not to use the `Environment` itself. That choice is debatable, but as it is a fundamental design decision, if it is debated it should be done on its own right.

Now, regarding how to handle the DPI scaling of the pointer. I don't think hard-coding a dpi factor of 4 and letting the compositor handle everything is the best approach. There have been discussions about other toolkits wrt to managing pointers, and it appears that the memory consumption of loading entire pointer themes with hi dpi scaling factors can actually be not that negligible.

I agree with your initial choice of using the max of the dpi factors of all outputs. I don't think tracking which output the pointer is on is the best approach, as we don't want to be reloading pointer themes all the time. To do that adding a new constructor to the `ThemeManager` accessing the `Environment` is indeed necessary, but I'd rather keep the old constructor as well for the reason detailed above. In this case, the old constructor would simply build a non-dpi-aware theme manager.

dvc94ch:
I have a couple of questions:

1. I noticed that when AutoThemer isn't used the cursor works just fine, at least in sway. Why is AutoThemer the default for every window implementation? It seems to only be required if you want to render your own cursor.

2. Can we expose the ShellSurface abstraction? A gui toolkit may want to render it's own decorations using some graphics api (opengl, vulkan).

dvc94ch:
More like moving the shell subfolder to a root folder and exposing a create_shell_surface from the Environment.

I'm have a couple of ideas I want to try out.
1. push back against the "webification" of all applications by creating a framework that bundles a wayland compositor instead of a webbrowser with the application for portability.

2. allow composing multiple applications into one. as an example the openscad gui has a text editor and a 3d model viewer. this creates a subpar experience, since the text editor is obviously crap. why not have text editor writers build awesome text editors then bundle the text editor with a 3d model viewer and some glue code using the framework described in 1.

So I basically don't need any window decorations.

vberger:
I see. Well it could make sense to expose the shell module this way I guess, it'd only require properly documenting all the API that'd get exposed.

PR #51, opened 2018-12-09
Expose shell interface.

dvc94ch:
Exposes the shell interface as discussed in #49


src/shell/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+mod wl;
+mod xdg;
+mod zxdg;
+
+/// Possible events generated by a window that you need to handle
+#[derive(Clone, Debug)]
+pub enum Event {
+    /// The state of your window has been changed
+    Configure {
+        /// Optional new size for your *inner* surface
```
vberger: This is no longer true. The events generated by the shell surfaces are the new size for the *whole* surface, including all decorations and everything. It is the `Window` abstraction that processes it to subtract the decoration size.

Also considering that the `Refresh` does not exist for a bare shell surface, I think it'd be better to create an `Event` enum specifically for the shell surface exposition, rather than re-use the `Window` one.

src/shell/mod.rs
```diff
@@ -0,0 +1,88 @@
+//! Unified wayland and xdg shell surface handling
```
vberger:
```suggestion
//! Unified shell surface handling
//!
//! This module provides abstraction unifying the various iterations of the shell surface
//! protocols (`wl_shell`, `zxdg_shell_v6` and `xdg_shell`, the current standard).
//!
//! This abstraction only manages the protocol part of shell surfaces. If you're looking
//! for a more battery-included abstraction for creating windows, consider the `Window`
//! type.
```

src/shell/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    Impl: FnMut(Event) + Send + 'static,
+{
+    match *shell {
+        Shell::Wl(ref shell) => Box::new(wl::Wl::create(surface, shell, implem)) as Box<_>,
+        Shell::Xdg(ref shell) => Box::new(xdg::Xdg::create(surface, shell, implem)) as Box<_>,
+        Shell::Zxdg(ref shell) => Box::new(zxdg::Zxdg::create(surface, shell, implem)) as Box<_>,
+    }
+}
+
+/// ShellSurface trait
```
vberger:
```suggestion
/// Trait abstracting over shell surface protocols
///
/// This trait's API is designed to reflect the behavior of the current standard shell
/// surface protocol: `xdg_shell`. Compatibility implementations are provided for older
/// protocols.
```

src/shell/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn unset_maximized(&self);
+    /// Minimize surface
+    fn set_minimized(&self);
+    /// Set geometry
+    fn set_geometry(&self, x: i32, y: i32, width: i32, height: i32);
+    /// Set minimum surface size
+    fn set_min_size(&self, size: Option<(i32, i32)>);
+    /// Set maximum surface size
+    fn set_max_size(&self, size: Option<(i32, i32)>);
+    /// Get xdg
```
vberger:
```suggestion
    /// Retrieve the `XdgToplevel` proxy if the underlying shell surface uses the `xdg_shell` protocol
    ///
    /// This allows interactions with other protocol extensions, like `xdg_decorations` for example.
```

src/env.rs
```diff
@@ -234,4 +236,16 @@ impl Environment {
         self.surfaces.lock().unwrap().push(surface.clone());
         surface
     }
+
+    /// Create a new shell surface
+    pub fn create_shell_surface<Impl>(
+        &self,
+        surface: &Proxy<wl_surface::WlSurface>,
+        shell_impl: Impl,
+    ) -> Arc<Box<ShellSurface>>
```
vberger: I don't think the wrapping into an `Arc` needs to be done at this level, this was an implementation detail of `Window`.

vberger:
Okay, this looks good, please add an entry in the changelog though.

PR #52, opened 2018-12-09
Cursor scaling

dvc94ch:
Implements dpi aware cursor loading.

* The theme manager api has changed, but that doesn't affect the andrew or winit crates (the only ones published on crates.io) I also removed a bunch of code that isn't used anywhere in the crates.io universe. This can be avoided if there is really a reason to keep it around.

* The frame api has a init_from_environment function added to it with a stub to use the init implementation. This disables any cursor theaming.

* The window api changes slightly, which IMO isn't a big deal, the andrew and winit crates should be impacted minimally, but it avoids maintaining a bunch of code we don't need.

### DIFF
--- a/src/surface.rs
+++ b/src/surface.rs
@@ -1,9 +1,8 @@
 //! Utility functions for creating dpi aware wayland surfaces.
-use env::Environment;
 use output::OutputMgr;
-use std::sync::Mutex;
+use std::sync::{Arc, Mutex};
 use wayland_client::protocol::wl_compositor::RequestsTrait as CompositorRequest;
-use wayland_client::protocol::{wl_output, wl_surface};
+use wayland_client::protocol::{wl_compositor, wl_output, wl_surface};
 use wayland_client::Proxy;
 
 pub(crate) struct SurfaceUserData {
@@ -67,14 +66,14 @@ impl SurfaceUserData {
 /// data. When the dpi value is updated the caller is notified through the
 /// dpi_change closure.
 pub(crate) fn create_surface<F>(
-    environment: &Environment,
+    compositor: &Proxy<wl_compositor::WlCompositor>,
+    output_manager: &OutputMgr,
     dpi_change: Box<F>,
 ) -> Proxy<wl_surface::WlSurface>
 where
     F: FnMut(i32, Proxy<wl_surface::WlSurface>) + Send + 'static,
 {
-    environment
-        .compositor
+    compositor
         .create_surface(move |surface| {
             surface.implement(
                 move |event, surface| {
@@ -93,7 +92,7 @@ where
                     };
                 },
                 Mutex::new(SurfaceUserData::new(
-                    environment.outputs.clone(),
+                    output_manager.clone(),
                     dpi_change,
                 )),
             )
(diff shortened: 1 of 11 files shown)


dvc94ch:
One other suggestion would be to move the decoration drawing stuff into it's own subfolder. The window can reexport BasicFrame and ConceptFrame, this keeps the purpose of the window to tying all the stuff together.

vberger:
I'm not very fan of all modules cloning the entire environment when they could only clone the few globals the need to keep around. And in this vein, I'm not 100% confident making the Environment clonable won't be a problem in the future as new functionality is added.

dvc94ch:
Yeah I agree. So I reworked the patches based on your comment by introducing a SurfaceManager that keeps track of surfaces. To create a surface you need a SurfaceManager and a Proxy<WlCompositor> and the Environment exposes a create_surface using it's Proxy<WlCompositor>

src/surface.rs
```diff
@@ -121,3 +120,53 @@ pub fn get_outputs(surface: &Proxy<wl_surface::WlSurface>) -> Vec<Proxy<wl_outpu
         .outputs
         .clone()
 }
+
+/// Surface Manager
+#[derive(Clone)]
+pub struct SurfaceManager {
+    output_manager: OutputMgr,
+    surfaces: Arc<Mutex<Vec<Proxy<wl_surface::WlSurface>>>>,
+}
```
vberger: I think at this point the `SurfaceManager` could also own a `Proxy<WlCompositor>`. This would allow to remove the need for passing it as argument to the `create_surface` method. The `wl_compositor` global is not really something that is supposed to change during the lifetime of the client.

dvc94ch: The problem is that the SurfaceManager needs to be created before the GlobalManager which is needed to get the WlCompositor.

vberger: Ah right. The usual chicken and egg problem...

Can't we work around that though? The `OutputMgr` cannot be created outside of the `Environment`, it may be acceptable for the `SurfacesMgr` to be the same way. This would, there could be a tighter coupling between its internals and the `Environment`, no?

dvc94ch:
I've been using `smithay-client-toolkit` for a bit now and there are a couple of things that I think are suboptimal. I experimented with a redesign that is `high dpi`, `multi-seat` and `multi-surface` from the start. Most notably it doesn't support window decorations yet and it uses a queue based api instead of a callback based one. If you like I'd appreciate your opinion on it's api. It's much closer to winit without abstracting any wayland details away.

Cheers David

- [0] [https://github.com/dvc94ch/linux-toolkit](https://github.com/dvc94ch/linux-toolkit)

vberger:
Sorry if you felt resistance wrt improving SCTK, I've had quite a lot on my plate during these last few weeks, this will remain the case for a few more weeks so I won't be able to really dig deep into serious design questions.

Now, I've glanced over your `linux-toolkit`, this is definitely an interesting work. I've been thinking about introducing something akin your `EventDrain` in SCTK, but got lost in trying to find the "absolute-perfect-design" before doing so.

For the record, an earlier version of `wayland-client` represented event queues as iterators, but I changed back to callback for several reasons. The primary one is that an iterator-based API could absolutely not work for `wayland-server`, and I wanted to share as much code as possible between client and server for the code-generation with `wayland-scanner`.

However, I had encountered other frictions with an iterator-based API, I don't know if you've already considered or solved them, but here they are if that can help you:

- First, the aggregation of events from multiple sources. For example if there are different seats and each with its own pointer, the app needs to distinguish events coming from both pointers from each other. Similarly, an app that creates several toplevel surfaces will need to be able to distinguish events from all of these surfaces.
  - The only two ways I saw to handle this are either tagging each event by joining it with a `Proxy<_>` of the object that generated it, or having each object of interest dump its events in a different event drain. Both of which have ergonomic drawbacks IMO.
- Second, input objects, notably touch, pointers and toplevel surfaces during an interactive resize, tend to generate a really big stream of events, and storing all of them in a shared queue before interpreting them (which often consists on only using the most recent anyway) may be a non-negligible resource cost. I don't know how important it is though.

In any case, if you find an API design that works well and are willing to collaborate it back into SCTK, and go forward together from it, that'll always be welcome. I'm by no means deeply attached to the current design. For comparison, I completely rewrote `wayland-rs` 4 times before reaching the current design that I'm more or less satisfied with...

dvc94ch:
It's not that I felt resistance, just thought that I could get there faster by making changes freely. I also underestimated the effort a little. 

In linux-toolkit every surface has an EventDrain and every input has a SeatEvenSource passed to it that knows it's seat id, wraps events in a SeatEvent and pushes it to the currently active surface's event queue.

One problem I've encountered is with key repeat. The event is queued on the repeat thread, but the events aren't processed until the next key event because the main thread is blocked with event_queue.dispatch(). Replacing event_queue.dispatch() with event_queue.dispatch_pending() doesn't work as I'd expect. I haven't spent much time trying to figure out why yet.

Another thing I'd like to try is replacing the event drain with a tokio stream. There is also an async unix socket implementation in tokio that wayland-rs could use. This would solve the second issue you mentioned by not needing intermediate event queues.

dvc94ch:
> The primary one is that an iterator-based API could absolutely not work for wayland-server

Can you elaborate on why it absolutely could not work? Is there anything inherently synchronous about the wayland protocol? I noticed some weirdness with clipboard handling as the events need to be processed between `display.flush()` and `event_queue.dispatch()`

vberger:
Well, as per the protocol, the wayland compositor is required to process client requests in order. So it is not an option to have different event queues server-side.

Even client-side with event queues, there is some event processing that needs to remain global actually, as the state of the protocol itself is pretty synchronous, notably regarding the assignment of object ids.

Issue #54, opened 2018-12-26
Build failure on big endian targets


io12:
For example, here is a build failure on `mips-unknown-linux-gnu`:

```
error[E0308]: mismatched types
  --> /cargo/registry/src/github.com-1ecc6299db9ec823/smithay-client-toolkit-0.4.3/src/window/basic_frame.rs:40:19
   |
40 |           if active {
   |  ___________________^
41 | |             #[cfg(target_endian = "little")]
42 | |             return [0x80, 0x80, 0x80, 0xFF];
43 | |             #[cfg(target_endian = "big")]
44 | |             [0xFF, 0x80, 0x80, 0x80];
   | |                                     - help: consider removing this semicolon
45 | |         } else {
   | |_________^ expected array of 4 elements, found ()
   |
   = note: expected type `[u8; 4]`
              found type `()`
                                                                                
error[E0308]: mismatched types
  --> /cargo/registry/src/github.com-1ecc6299db9ec823/smithay-client-toolkit-0.4.3/src/window/basic_frame.rs:45:16
   |
45 |           } else {
   |  ________________^
46 | |             #[cfg(target_endian = "little")]
47 | |             return [0x60, 0x60, 0x60, 0xFF];
48 | |             #[cfg(target_endian = "big")]
49 | |             [0xFF, 0x60, 0x60, 0x60];
   | |                                     - help: consider removing this semicolon
50 | |         }
   | |_________^ expected array of 4 elements, found ()
   |
   = note: expected type `[u8; 4]`
              found type `()`
                                                                                
error[E0308]: mismatched types
  --> /cargo/registry/src/github.com-1ecc6299db9ec823/smithay-client-toolkit-0.4.3/src/window/basic_frame.rs:89:38
   |
89 |               ButtonState::Disabled => {
   |  ______________________________________^
90 | |                 #[cfg(target_endian = "little")]
91 | |                 return [0x20, 0x80, 0x80, 0xFF];
92 | |                 #[cfg(target_endian = "big")]
93 | |                 [0xFF, 0x80, 0x80, 0x20];
   | |                                         - help: consider removing this semicolon
94 | |             }
   | |_____________^ expected array of 4 elements, found ()
   |
   = note: expected type `[u8; 4]`
              found type `()`
                                                                                
error[E0308]: mismatched types
  --> /cargo/registry/src/github.com-1ecc6299db9ec823/smithay-client-toolkit-0.4.3/src/window/concept_frame.rs:40:19
   |
[...]

PR #55, opened 2018-12-26
Fix build error on big-endian targets

io12:
Fixes issue https://github.com/Smithay/client-toolkit/issues/54

### DIFF
--- a/src/window/basic_frame.rs
+++ b/src/window/basic_frame.rs
@@ -38,15 +38,17 @@ impl Theme for DefaultTheme {
     // Used for header color
     fn get_primary_color(&self, active: bool) -> [u8; 4] {
         if active {
-            #[cfg(target_endian = "little")]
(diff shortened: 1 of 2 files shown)


dvc94ch:
Why don't you fix it like the error message suggests by removing the semicolon?

io12:
I figured it would be better for it to be an `if` expression so the compiler can check both branches (to prevent an issue like this from happening again). Also, it's inconsistent to have a return statement on one branch and an implicit return on another.

vberger:
> I figured it would be better for it to be an if expression so the compiler can check both branches (to prevent an issue like this from happening again).

I agree, this `if` will be eliminated by the optimizer anyway, so it's better this way. :+1:

vberger:
Thanks!

io12:
Can a new release be created so this fix is available?

vberger:
v0.4.4 is now released

io12:
Thanks!

Issue #56, opened 2019-01-16
Cannot `set_title()` while handling window events


trimental:
When using `set_title()` and also handling an event of the windows implementation a mutex deadlock occurs. 

`set_title()` triggers a `FrameRequest::Refresh` by locking the windows implementation. If the windows implementation is locked while inside the windows implementation then a deadlock occurs.

A small example being
```
match next_action.lock().unwrap().take() {
  Some(WEvent::Refresh) => {
    window.set_title("refreshed".into());
    window.refresh();
    window.surface().commit();
  }
}
```

vberger:
Indeed, there is some inconsistency between some methods of window. Some require a refresh to be called afterwards, and some trigger a refresh event directly, I think we could harmonize all that.

PR #57, opened 2019-01-16
Require a `refresh()` after `set_title()` for changes to take effect

trimental:
The automatic triggering of a refresh after calling `set_title()` when also handling a window event causes a mutex deadlock. 

Requiring the user to manually refresh the window after setting the title fixes this issue.

I cannot find any other instances where this sort of deadlock would occur but would be happy to include them in this PR if there are.

Fixes https://github.com/Smithay/client-toolkit/issues/56

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


vberger:
Thanks!

PR #61, opened 2019-01-26
Use `subsec_nanos()` of time since UNIX_EPOCH for random shm file han…

trimental:
Removes the rand crate by instead using the time since UNIX_EPOCH's subsecond nanoseconds. Its good because its a fairly random number that changes every nanosecond and if we somehow get two of the same handles then we can just try again a nanosecond later.


vberger:
Merged as https://github.com/Smithay/client-toolkit/commit/116f45454b2b671506b8d838efcf1f16ce61b2a8

PR #62, opened 2019-01-27
Fix system repeat rate

trimental:
This was a mistake I made when implementing the key repetition. The delay should be milliseconds but the repeat rate is repeats per second.

```
from wayland docs:
rate | the rate of repeating keys in characters per second
delay | delay in milliseconds since key down until repeating starts
```

Thanks to @e00E for catching this bug

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Keyboard: fix system repeat rate as repeats per second rather then millisecond delay between repeats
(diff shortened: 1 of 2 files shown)


e00E:
That looks like the same place I identified as well. Should also change documentation at line 364 to `in characters per second` instead of `in milliseconds`.

trimental:
@e00E good spot, thanks

vberger:
Thanks!

Issue #67, opened 2019-03-19
Button (inactive) colors doesn't appear to do anything


JMurph2015:
This is coming from from much higher in the stack, but in my testing on Alacritty, which uses this library via `winit`, changing the button color (other than the hovered color) doesn't seem to do anything.  Can anyone observe similar results?

vberger:
This is actually a characteristic of the `ConceptFrame`, which does not color inactive buttons. If you check the `BasicFrame` for example, the inactive color is use when the buttons are not hovered.

JMurph2015:
Changed in #68, buttons backgrounds are now shaded on `ConceptFrame`.

PR #68, opened 2019-03-19
Added button icon color to the Theme trait and used it in the ConceptFrame

JMurph2015:
This would allow clients to make the ConceptFrame's buttons react to being hovered over (and also customize their foreground for each button individually).  A default implementation is provided to maintain compatibility, and it provides the same behavior as existing.


JMurph2015:
Woops, this also modifies ConceptFrame to paint the `ButtonState::Idle` color for each button as shading around the icon.  This changes the behavior of existing clients, but I don't see why this shouldn't be exposed through that mechanism (Fixes #67).

vberger:
I see two changes in this PR, the first is to make `ConceptFrame` actually use the inactive colors of its buttons and display them. The aesthetic gain from than can be discussed, but why not after all.

I'm quite skeptical about the addition of of the icon-specific colors though. The themes can already compute new colors internally by blending the primary/secondary colors and the button colors for example. Is there really a need to add yet another configuration color?

It feels like you're trying to change the `ConceptFrame` theme while the solution to your problem would actually be to create a new theme the behaves like you want (and maybe push winit to use it ?).

The idea of SCTK is really that a few simple themes are provided out of the box, and SCTK's users can create their own themes if they wish so.

JMurph2015:
The concerns you have are pretty fair.  I just thought it would be nice to be able to control the button stroke color so that clients could have the stroke change on hover as well.  This is nice for situations when the secondary color would look bad against the hover button color.  As for the button background color situation, I saw that as pretty unintuitive as it was, and this behavior makes a lot more sense.

vberger:
I see. In that case, could you clarify in the trait documentation of the `_icon_` methods that if they are not overridden, they will default to the active theme secondary color ?

Also, it's nice to have introduced an example to showcase theme customization, but I don't think it's needed to keep the "keyboard input handling" logic in it, it is of-topic for this example.

vberger:
Finally, your PR will also need a rustfmt run for the CI to pass. :)

vberger:
Okay, this looks good to me :+1: 

Could you just add a changelog entry for this ? Thanks.

vberger:
Thanks!

PR #76, opened 2019-07-05
Add confine pointer

### LINKED ISSUE #72: Support for pointer grab
We need some abstraction for pointer grab mechanisms, exposing a convenient API for the [pointer-constraints](https://docs.rs/wayland-protocols/0.23.5/wayland_protocols/unstable/pointer_constraints/index.html) and [relative-pointer](https://docs.rs/wayland-protocols/0.23.5/wayland_protocols/unstable/relative_pointer/index.html) protocols.

Cherser-s:
Resolves #72. It doesn't look good right now, and I didn't add the ability to grab pointer for new seats that can be added after pointer grab yet, since it requires providing `new_seat` method with surface reference.

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -11,6 +11,7 @@ use wayland_client::protocol::{
 use wayland_client::{EventQueue, GlobalEvent, GlobalManager, NewProxy};
 use wayland_protocols::unstable::xdg_decoration::v1::client::zxdg_decoration_manager_v1;
 use wayland_protocols::unstable::xdg_shell::v6::client::zxdg_shell_v6;
+use wayland_protocols::unstable::pointer_constraints::v1::client::zwp_pointer_constraints_v1;
 use wayland_protocols::xdg_shell::client::xdg_wm_base;
 
 /// Possible shell globals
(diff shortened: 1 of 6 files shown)


vberger:
Hi, thanks for working on this!

Before all, could you explain me why you merged this logic with the `Frame` implementations, rather than creating a new module with this specific logic? Is there some coupling between them?

Cherser-s:
Well because frames (and not window module) are holding references to WlPointer right now, so I thought that would be more convenient. 
So I think to add some kind of `get_pointers` method to `Frame` module instead. (still requires updating pointer grab on new seat in case if new seat is created)
Also, I don't know whether it's better to constrain all available seats to surface at once, or constrain only specified one.

vberger:
A SCTK user would normally also have a `WlPointer` at hand without using the one internal to the Window/Frames, and separating the two would make for a more flexible API overall I think: users may wish to confine the pointer to some other surface than the one backing a Window.

In this approach, I guess we could have some API allowing the app to constrain a given pointer to a given surface, something like that.

Cherser-s:
Ok, so `Environment` should contain an module for pointer grab (which contains a `pointer_constraints` object from the start and methods for both pointer confinement and lock).

vberger:
I think ideally this code should be separated into its own module (like the `data_device`, `keyboard` or `pointer` modules). Though this does raise some questions of API design for SCTK I had in mind for some time. Will make an issue about that.

Cherser-s:
In this case there is no need to even implement API for that purpose since all you need to do is call one method (whether it is `lock_pointer` or `confine_pointer`). Client does only need to hold reference to created object to destroy it if needed.

vberger:
I'm going to close this, given this appear to need more design thought. Please head to #77 if you want to discuss this further. :)

PR #84, opened 2019-10-13
do not crash when fonts don't have extensions

crepererum:
This happened (for me) when using alacritty on a wayland desktop. Seems
that there is some messed up font in the fontconfig list. Let's just
ignore these broken fonts.

### DIFF
--- a/src/window/concept_frame.rs
+++ b/src/window/concept_frame.rs
@@ -484,12 +484,9 @@ impl Frame for ConceptFrame {
                                 .get_regular_family_fonts("sans")
                                 .unwrap()
                                 .iter()
-                                .filter_map(|p| {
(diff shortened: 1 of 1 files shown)


src/window/concept_frame.rs
```diff
@@ -484,12 +484,15 @@ impl Frame for ConceptFrame {
...
+                                .filter_map(|p| match p.extension() {
+                                    Some(e) => {
+                                        if e == "ttf" {
+                                            Some(p)
+                                        } else {
+                                            None
+                                        }
                                     }
+                                    None => None,
                                 })
```
vberger: The `match` block can compacted using `Some(e) if e == "ttf" => { ... }`

PR #85, opened 2020-01-15
Deep rework of the whole API

vberger:
This is a large rework of SCTK's API around the following principles:

- Update to soon to be wayland-client 0.25 (actually to master, which contains more fixes)
- Integration with calloop (master version likewise)
- The whole crate is now built in a monolithic-modular way, around the `Environment<_>` type, which is generated by a macro allowing users to introduce their own modules into the architecture.
- bookkeeping is as much as possible done by SCTK (tracking seats for example), making the modules more tightly coupled together
- integration of the `DispatchData` mechanism from wayland-client into all SCTK callbacks, allowing the app to access its global state from everywhere
- secondary sources of events (key repetition and data offer reading currently) are also integrated into calloop

This is the big stepping stone kickstarting the API rework of SCTK, more stuff will need to be done to further improve it.


kchibisov:
It seems like the key repeat is way slower on `kbd_input` example compared to the current master, not sure if you're aware of it or not.

kchibisov:
I don't have much time right now to find it where you're handling it, but you should invert the time. it should be `1000ms / repeat_rate` and not just `repeat_rate`. So on your PR the lower `repeat_rate` in compositor you have the faster it'll be in `kbd_input` example and vise-versa.

vberger:
Ah, right, thanks for the heads-up. I forgot that the repeat rate in protocol is expressed in `Hz`, and not in `ms`.

PR #94, opened 2020-02-03
Fix `set_decorations` arbitrary changing SSD mode to CSD

kchibisov:
SCTK by default uses SSD decorations, however `set_decorations` true,
could force it to draw decorations and unset_mode assuming it's SSD,
however compositor can change SSD to CSD, so this unset will fallback
back to SSD and it'll mess up inner state.

This commit makes `set_decorations` to hide/show decorations only in CSD
mode, in SSD mode it's up to server on what to do.

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -139,7 +139,6 @@ pub struct Window<F: Frame> {
     frame: Arc<Mutex<F>>,
     surface: wl_surface::WlSurface,
     decoration: Mutex<Option<zxdg_toplevel_decoration_v1::ZxdgToplevelDecorationV1>>,
-    decoration_mode: Arc<Mutex<Option<DecoratoinsMode>>>,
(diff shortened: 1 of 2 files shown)


kchibisov:
I've tested in alacritty + winit on GNOME,weston and sway(with going from SSD to CSD, and then toggling decorations). I was toggling decorations with from `set_fullscreen` in winit, with `set_decorate(false/true)`.

src/window/mod.rs
```diff
@@ -12,6 +12,8 @@ use wayland_protocols::unstable::xdg_decoration::v1::client::{
     zxdg_decoration_manager_v1, zxdg_toplevel_decoration_v1,
 };
 
+use self::zxdg_toplevel_decoration_v1::{Mode as DecoratoinsMode};
```
YaLTeR: Typo: Decorat**oi**nsMode

PR #95, opened 2020-02-05
HiDPI scaling for decorations

kchibisov:
picture

I've also fixed an issue with most left-bottom most pixel in a different color and line separator not drawing.

### DIFF
--- a/src/surface.rs
+++ b/src/surface.rs
@@ -107,11 +107,11 @@ pub fn get_dpi_factor(surface: &wl_surface::WlSurface) -> i32 {
 
 /// Returns the current dpi factor of a surface if it was created by SCTK or falling back to 1 if
 /// it's not.
-pub(crate) fn get_dpi_factor_or_fallback(surface: &wl_surface::WlSurface) -> i32 {
+pub(crate) fn try_get_dpi_factor(surface: &wl_surface::WlSurface) -> Option<i32> {
     if let Some(user_data) = surface.as_ref().user_data::<Mutex<SurfaceUserData>>() {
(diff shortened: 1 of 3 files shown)


src/window/concept_frame.rs
```diff
@@ -804,18 +827,25 @@ fn request_for_location(
...
-    );
-    canvas.draw(&division_line);
+
+    for i in 1..=scale {
+        let y = HEADER_SIZE as usize * scale - i;
+        let division_line = line::Line::new(
+            (0, y),
+            (width as usize * scale, y),
+            theme.get_secondary_color(false),
+            false,
```
kchibisov: antialias `true` was breaking things here, since it was impossible to draw a horizontal line with it ((0,y),(x,y)). it should be drawn as ((0,y),(x,y - 1)) before, so I've disabled it, since it doesn't make much sense for horizontal line anyway.

src/surface.rs
```diff
@@ -105,6 +105,16 @@ pub fn get_dpi_factor(surface: &wl_surface::WlSurface) -> i32 {
         .dpi_factor
 }
 
+/// Returns the current dpi factor of a surface if it was created by SCTK or falling back to 1 if
+/// it's not.
+pub(crate) fn get_dpi_factor_or_fallback(surface: &wl_surface::WlSurface) -> i32 {
```
kchibisov: we can probably name it `try_get_dpi_factor` and make it return `Some(i32)` and handle unwrap from a caller.

vberger: Sound good :+1:

PR #97, opened 2020-02-07
Fix HiDPI cursor icon position [backport]

kchibisov:
It seems like just porting the same code from 0.7.0 didn't work. While updating winit, I've noticed that cursor position reported is a bit different compared to shape. I don't quite get why this fixes the issue, however I think I should normalize thing?

I wonder why I can't notice it on v0.7.0 tbh, but maybe something was change internally?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Fix `HiDPI` cursor bounding box
+
 ## 0.6.5 -- 2020-02-07
 
 #### Bugfixes
--- a/src/pointer/theme.rs
+++ b/src/pointer/theme.rs
@@ -231,7 +231,7 @@ impl ThemedPointer {
         } else {
             // surface is old and does not support damage_buffer, so we damage
             // in surface coordinates and hope it is not rescaled
-            surface.damage(0, 0, w, h);
+            surface.damage(0, 0, w / scale, h / scale);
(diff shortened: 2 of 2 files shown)


kchibisov:
Ok, After extensive investigation this bug was reproduced on the SCTK master, so it also requires a fix. I didn't notice it because new version don't reposition the cursor for me, etc. You can easily repro it if you with `kbd_input` example and doing the following transformation to `ConceptFrame`.

```
diff --git a/src/window/concept_frame.rs b/src/window/concept_frame.rs
index 296da34..ae2e467 100644
--- a/src/window/concept_frame.rs
+++ b/src/window/concept_frame.rs
@@ -767,7 +767,7 @@ fn change_pointer(pointer: &AutoPointer, location: Location, serial: Option<u32>
         Location::BottomLeft => "bottom_left_corner",
         Location::Left => "left_side",
         Location::TopLeft => "top_left_corner",
-        _ => "left_ptr",
+        _ => "text",
     };
     if pointer.set_cursor(name, serial).is_err() {
         log::error!("Failed to set cursor");

```

And try to move your cursor to a decorations button. You can see a difference between very easily.

This is an activation point on HiDPI monitor(scale 2).

pic

But this is on scale 1.

pic

Point close to it on scale 2

pic


As you can see on a last picture you it doesn't highlight, because the actual cursor is lower.

PR #99, opened 2020-02-08
Fix HiDPI cursor icon position

kchibisov:
For more see https://github.com/Smithay/client-toolkit/pull/97#issuecomment-583773503 .

### DIFF
--- a/src/seat/pointer/theme.rs
+++ b/src/seat/pointer/theme.rs
@@ -207,16 +207,15 @@ struct PointerInner {
 impl PointerInner {
     fn update_cursor(&self, pointer: &wl_pointer::WlPointer) -> Result<(), ()> {
         let mut themes = self.themes.borrow_mut();
-        let cursor = themes
-            .get_cursor(&self.current_cursor, self.scale_factor as u32)
-            .ok_or(())?;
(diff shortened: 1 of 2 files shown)


src/seat/pointer/theme.rs
```diff
@@ -207,16 +207,15 @@ struct PointerInner {
...
             .frame_info(0)
-            .map(|(w, h, hx, hy, _)| (w as i32, h as i32, hx as i32, hy as i32))
+            .map(|(w, h, hx, hy, _)| (w as i32, h as i32, (hx / scale) as i32, (hy / scale) as i32))
             .unwrap_or((0, 0, 0, 0));
 
-        self.surface.set_buffer_scale(self.scale_factor);
+        self.surface.set_buffer_scale(scale as i32);
         self.surface.attach(Some(&buffer), 0, 0);
         if self.surface.as_ref().version() >= 4 {
             self.surface.damage_buffer(0, 0, w, h);
```
vberger: This line and the one below (`self.surface.damage`) cannot both be using the same `w` and `h`. `damage_buffer` requires coordinates in buffer space (aka "physical" pixels), while `damage` requires coordinates in surface space (aka "logical" pixels).

PR #101, opened 2020-02-10
Add HiDPI scaling for decorations

### LINKED ISSUE #90: Left bottom corner pixel in CSD has a different color
picture

### LINKED ISSUE #91: Hi-DPI aware decorations
The `ConceptFrame` currently provided with SCTK is not dpi-aware and could easily be.

kchibisov:
Fixes #91.
Fixes #90.

### DIFF
--- a/src/window/concept_frame.rs
+++ b/src/window/concept_frame.rs
@@ -112,12 +112,33 @@ impl Part {
         parent: &wl_surface::WlSurface,
         compositor: &Attached<wl_compositor::WlCompositor>,
         subcompositor: &Attached<wl_subcompositor::WlSubcompositor>,
+        inner: Option<Rc<RefCell<Inner>>>,
     ) -> Part {
-        let surface = crate::surface::setup_surface(
-            compositor.create_surface(),
-            None::<fn(_, _, DispatchData)>,
-        );
+        let surface = if let Some(inner) = inner {
+            crate::surface::setup_surface(
+                compositor.create_surface(),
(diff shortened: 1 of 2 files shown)


src/window/concept_frame.rs
```diff
@@ -112,12 +112,33 @@ impl Part {
...
+            crate::surface::setup_surface(
+                compositor.create_surface(),
+                Some(
+                    move |dpi, surface: wl_surface::WlSurface, _ddata: DispatchData| {
+                        surface.set_buffer_scale(dpi);
+                        surface.commit();
+                    },
+                ),
+            )
+        };
```
kchibisov: I'm not sure if this is something desirable tbh. I'm not an expert in plumbing and sctk internals, so suggestions are welcome.

vberger:
I'll keep this waiting until we get to the bottom of #102, given the issue was introduced by #95

kchibisov:
@vberger the issue wasn't introduced by #95 it became more visible due to it. Previously the parallel line was drawn in a bit odd way due to anti-aliasing. In fact on #90 this black pixel is a separator line. I can send you a picture from old SCTK before #95, you'll see the same blue lines.

src/window/concept_frame.rs
vberger: I'm alright with that.

vberger:
Okay, so this will need to be rebased, and then I'm good with merging it.

PR #108, opened 2020-03-15
Implement listen_for_outputs

ifreund:
This function, available through the OutputHandling trait, makes it
possible to register a listener for the creation and removal of outputs.

This is as of yet untested as I wasn't quite sure what the best practice is there.  Should I just create a simple example using it?

### DIFF
--- a/examples/layer_shell.rs
+++ b/examples/layer_shell.rs
@@ -1,0 +1,255 @@
+use smithay_client_toolkit::{
+    environment,
+    environment::{Environment, SimpleGlobal},
+    output::{with_output_info, OutputHandler, OutputHandling, OutputInfo, OutputStatusListener},
+    reexports::{
+        calloop,
+        client::protocol::{wl_compositor, wl_output, wl_shm, wl_surface},
+        client::{Attached, Display, Main, Proxy},
+        protocols::wlr::unstable::layer_shell::v1::client::{
+            zwlr_layer_shell_v1, zwlr_layer_surface_v1,
+        },
+    },
+    shm::{DoubleMemPool, ShmHandler},
+    WaylandSource,
+};
+
+use byteorder::{NativeEndian, WriteBytesExt};
+
+use std::cell::{Cell, RefCell};
+use std::io::{BufWriter, Seek, SeekFrom, Write};
+use std::rc::Rc;
+
+struct Env {
+    compositor: SimpleGlobal<wl_compositor::WlCompositor>,
+    layer_shell: SimpleGlobal<zwlr_layer_shell_v1::ZwlrLayerShellV1>,
+    shm: ShmHandler,
+    outputs: OutputHandler,
+}
+
+impl OutputHandling for Env {
+    fn listen<F: FnMut(wl_output::WlOutput, &OutputInfo) + 'static>(
+        &mut self,
+        f: F,
+    ) -> OutputStatusListener {
+        self.outputs.listen(f)
+    }
+}
+
+environment!(Env,
+    singles = [
+        wl_compositor::WlCompositor => compositor,
+        zwlr_layer_shell_v1::ZwlrLayerShellV1 => layer_shell,
+        wl_shm::WlShm => shm,
+    ],
+    multis = [
+        wl_output::WlOutput => outputs
+    ]
+);
+
+#[derive(PartialEq, Copy, Clone)]
+enum RenderEvent {
+    Configure { width: u32, height: u32 },
+    Closed,
+}
+
+struct Surface {
+    surface: wl_surface::WlSurface,
+    layer_surface: Main<zwlr_layer_surface_v1::ZwlrLayerSurfaceV1>,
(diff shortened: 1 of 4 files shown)


vberger:
LGTM overall. I don't think this requires an example for merging, though an example that uses layer-shell to display a small thing on all screens to illustrate how to use layer-shell & track outputs would definitely be a positive addition!

Have you manually tested this on your computer, or not at all?

ifreund:
Alright, in implementing the `layer_shell` example I rethought a few things about my `listen_for_outputs` implementation. The callback now takes a `WlOutput` and a `&OutputInfo`, which matches the signature of the per-output callback option minus the `DispatchData`.  There is no reason to return an `Attached<WlOutput>` as the my initial implementation did since callbacks can be registered with `add_output_listener()`. Also, the `OutputStatus` enum I had previously added proved insufficient when testing with actual code, passing an `&OutputInfo` instead provides the same information and more as well as reducing overhead by removing the need to make a `with_output_info()` call.

The example I added crates a 32x32 pixel green box in the top left corner of all outputs, handling creation and removal of outputs through a listener.  I tested this in a nested sway session using `WLR_WL_OUTPUTS=2` and `swaymsg output enable/disable` to toggle outputs on/off and test the listening.

Would appreciate some code review on the example from the author of the library, I'd like to be sure Its not recommending any bad practices.

examples/layer_shell.rs
```diff
@@ -0,0 +1,255 @@
...
+environment!(Env,
+    singles = [
+        wl_compositor::WlCompositor => compositor,
+        zwlr_layer_shell_v1::ZwlrLayerShellV1 => layer_shell,
+        wl_shm::WlShm => shm,
+    ],
+    multis = [
+        wl_output::WlOutput => outputs
+    ]
+);
```
vberger: Unless I'm missing some subtlety, declaring the `Env` struct could be done using the `default_environment!()` macro with layer shell as an addon, like so:

```rust
default_environment!(Env,
    fields = [
        layer_shell: SimpleGlobal<zwlr_layer_shell_v1::ZwlrLayerShellV1>,
    ],
    singles = [
                 zwlr_layer_shell_v1::ZwlrLayerShellV1 => layer_shell
    ],
)
```

rather than manually creating the struct and its implementations.

examples/layer_shell.rs
```diff
@@ -0,0 +1,255 @@
...
+                self.dimensions = (width, height);
+                self.draw();
+                false
+            }
+            None => false,
+        }
+    }
+
+    fn draw(&mut self) {
+        let pool = self.pools.pool().unwrap();
```
vberger: I think it may be good to add some comment explaining that the `.unwrap()` here is used for the simplicity of the example. A real app should be able to handle `pool()` returning `None` (meaning the compositor is lagging behind and has not released the buffer soon enough).

examples/layer_shell.rs
```diff
@@ -0,0 +1,255 @@
...
+                shm: ShmHandler::new(),
+                outputs: OutputHandler::new(),
+            },
+        );
+        let ret = queue.sync_roundtrip(&mut (), |_, _, _| unreachable!());
+        ret.and_then(|_| queue.sync_roundtrip(&mut (), |_, _, _| unreachable!()))
+            .expect("Error during initial setup");
+
+        (env, display, queue)
+    };
```
vberger: Similarly, I think this can be done using `init_default_environment!`:

```rust
let (env, display, queue) = init_default_environment!(Env,
    fields=[
        layer_shell: SimpleGlobal::new()
    ]
).expect("Unable to connect to the wayland compositor");
```

src/output.rs
```diff
@@ -382,6 +424,51 @@ pub struct OutputListener {
...
+
+/// Trait representing the OutputHandler functions
+///
+/// Implementing this trait on your inner environment struct used with the
+/// [`environment!`](../macro.environment.html) by delegating it to its
+/// [`OutputHandler`](struct.OutputHandler.html) field will make available the output-associated
+/// method on your [`Environment`](../environment/struct.Environment.html).
+pub trait OutputHandling {
+    /// Insert a listener for output creation and removal events
+    fn listen<F: FnMut(WlOutput, &OutputInfo) + 'static>(&mut self, f: F) -> OutputStatusListener;
```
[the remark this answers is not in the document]
ifreund: Oversight I suppose, I'll add it

examples/layer_shell.rs
ifreund: Yeah that should be totally fine.

examples/layer_shell.rs
ifreund: For sure, I'll add a comment

ifreund:
Alright, I implemented all of your suggestions, thanks for the review!

vberger:
Alright, this is good! I just realized this is missing one last thing: a changelog entry. :sweat_smile:

ifreund:
Added!

PR #111, opened 2020-04-08
Make andrew and calloop into optional dependencies

vberger:
For users that want a leaner SCTK experience, these can bring a lot of
transitive dependencies that are not used in the end. This commit thus
introduces two cargo features:

- The `calloop` feature controls the integration with calloop. Disabling
  it removes the support for keyboard repetition and the provided
  calloop glue code for data device and the event queue.
- The `frames` feature controls the existence of implementations of the
  `Frame` trait (currently only `ConceptFrame`). These bring a lot of
  transitive dependencies via `andrew` which are not needed for
  libraries or app not needing decorations.


jplatte:
> * The `frames` feature controls the existence of implementations of the
>   `Frame` trait (currently only `ConceptFrame`). These bring a lot of
>   transitive dependencies via `andrew` which are not needed for
>   libraries or app not needing decorations.

This is perfect for use cases like the bar I've been working on, which currently uses the gtk stack but that I might port to smithay + some lower level webview than webkit2gtk if that turns out to be feasible.

PR #114, opened 2020-04-13
Add `seat_data()` method as a shorthand to get `SeatData`

kchibisov:
It's a bit nicer way than doing something like.

```rust
if let Some(seat_data) = seat::with_seat_data(&seat, |seat_data| seat_data.clone()) {

}
```

So it should be just
```rust
if let Some(seat_data) = seat::seat_data(&seat) {

}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+#### Additions
+
+- Add `clone_seat_data()` method as a shorthand to get `SeatData`
+
 ## 0.8.1 -- 2020-04-09
 
 #### Additions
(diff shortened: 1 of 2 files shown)


vberger:
Yes should be fine since `SeatData` is a pretty small thing.

Could you just clarify in the docs that this returns a copy of the `SeatData` (that will thus not be updated when the real `SeatData` is)? Maybe worth naming the function `clone_seat_data` or something like that?

kchibisov:
>Maybe worth naming the function clone_seat_data or something like that?

Maybe such naming actually makes sense, since it's `seat_data.clone()` for the reader.

vberger:
Perfect, thanks!

PR #117, opened 2020-04-19
Fix deadlock in surface scale factor callback

### LINKED ISSUE #116: get_surface_outputs deadlocks when called from create_surface_with_scale_callback
First locked in setup_surface{quick_assign} for enter/leave
The callback is called from recompute_scale_factor.
Suggestion:
Don't pass the callback to enter/leave (locked).
call recompute_scale_factor directly from the event handler, make it return new_scale_factor.
unlock
Call the callback

Matthias-Fauconneau:
Fixes #116

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Add `clone_seat_data()` method as a shorthand to get `SeatData`
 
+#### Bugfixes
+
+- Surface lock held across scale factor callback deadlocks scale factor API.
+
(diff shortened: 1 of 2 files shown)


vberger:
Okay, this looks good. Indeed not having the user_data locked when the user callback is invoked it better. :+1: 

Could you just add a Changelog entry regarding that?

Matthias-Fauconneau:
I don't know if the commit already notifies you ;)

vberger:
Perfect, thanks!

Issue #120, opened 2020-04-28
map_keyboard: Unused type parameter `Data`


Matthias-Fauconneau:
was for LoopHandle<Data> in map_keyboard_with_repeat which does not exists for map_keyboard

PR #121, opened 2020-04-28
Fix map_keyboard

Matthias-Fauconneau:
#120


vberger:
Could you please remove the second commit? This file is auto-generated using the `update-keysyms.sh` script, and so should not be modified manually. We should rather add the appropriate `allow` here:

https://github.com/Smithay/client-toolkit/blob/9d4b4b52df94771071d6904ba6cdd2236e5d0d89/update_keysyms.sh#L15

PR #125, opened 2020-05-03
Add primary selection helpers

kchibisov:
This commit adds helpers to handle `zwp_primary_selection` and
`gtk_primary_selection` via the common interface. The interface is
designed in a way, so user will use `zwp_primary_selection` by default
and fall back to `gtk_primary_selection` if the former is not available.


examples/selection.rs
```diff
@@ -143,6 +146,8 @@ fn main() {
             }
         }
     });
+    // Get primary selection manager
+    env.get_primary_selection_manager();
```
kchibisov, line 150: This is not something ideal, so I wonder how we should handle such init of primary selection, since it's done as LazyGlobal, like `Shell`, for `Shell` it's not a problem, since you're asking for configure/creating surface, but for selection it could be desired to initialize it transparently in a calls on a helper.

examples/selection.rs
```diff
@@ -63,7 +64,9 @@ fn main() {
             }
         })
         .expect("Failed to create a window !");
-    window.set_title("Selection".to_string());
+    window.set_title(
+        "Selection example, press c/C p/P to copy/paste from selection/primary clipboard respectively".to_string(),
+    );
```
vberger, line 69: That's a pretty long title, not sure if it'll always be readable. I'd rather print it to stdout, given the rest of the interactions of this example are already on stdout anyway.

examples/selection.rs
vberger: I'd suggest we add this call into the `init_default_environment!()` macro, and add a note in the module-level docs about the need to do that for users not using the macro.

PR #127, opened 2020-05-27
Maintenance/update dependencies

azriel91:
Updates dependency versions.

The main change here is `andrew` from `0.2.0` to `0.3.0`, which has a number of dependency upgrades (see [andrew#2/files](https://github.com/trimental/andrew/pull/2/files)), and raises the minimum supported Rust version to 1.41.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ nix = "0.17"
 dlib = "0.4"
 lazy_static = "1.0"
 memmap = "0.7"
-andrew = { version = "0.2.0", optional = true }
+andrew = { version = "0.3.0", optional = true }
 log = "0.4"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -11,18 +11,18 @@ keywords = ["wayland", "client"]
 description = "Toolkit for making client wayland applications."
 
 [dependencies]
-bitflags = "1.0"
+bitflags = "1.2"
```
vberger, line 14: There is no need to update minor versions of other dependencies like that, cargo already considers these versions as compatible with each other, so `cargo update` will already pull them.

azriel91: I've reverted these back and only updated `andrew`.

PR #130, opened 2020-05-30
remove unnecessary parentheses

PonasKovas:
eliminates 3 warnings.


kchibisov:
This file is auto generated, you should update the scripts, which generate it and regenerate the file. 


Also https://github.com/Smithay/client-toolkit/pull/121#issuecomment-620484766

vberger:
No, actually this file is *not* auto-generated, only `keysyms.rs` is, and its script have already been updated, I just forgot about `ffi.rs`.

@PonasKovas Thanks!

PR #131, opened 2020-05-30
Raqote (andrew replacement) [WIP]

trimental:
Here's a work in progress PR to use raqote instead of andrew. There's a few issues such as non native endian and screen flickering (probably due to not writing directly to the memory map) to address.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,9 @@ nix = "0.17"
 dlib = "0.4"
 lazy_static = "1.0"
 memmap = "0.7"
-andrew = { version = "0.3.0", optional = true }
+raqote = "0.8"
+font-kit = "0.7"
+pathfinder_geometry = "0.5"
 log = "0.4"
 wayland-client = "0.26"
 wayland-protocols = { version = "0.26" , features = ["client", "unstable_protocols"] }
@@ -25,8 +27,7 @@ calloop = { version = "0.6.1", optional = true }
 byteorder = "1.0"
 
 [features]
-default = ["frames", "calloop"]
-frames = ["andrew"]
+default = ["calloop"]
 
 [dev-dependencies]
 image = "0.23"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -16,7 +16,9 @@ nix = "0.17"
 dlib = "0.4"
 lazy_static = "1.0"
 memmap = "0.7"
-andrew = { version = "0.3.0", optional = true }
+raqote = "0.8"
+font-kit = "0.7"
```
kchibisov, line 20: Be ware of `font-kit`, it'll start linking to system libs, also font-kit isn't something that works on linux, since it's lacking font fallback and you won't be able to naturally handle certain things.

Cargo.toml
```diff
@@ -16,7 +16,9 @@ nix = "0.17"
 dlib = "0.4"
 lazy_static = "1.0"
 memmap = "0.7"
-andrew = { version = "0.3.0", optional = true }
+raqote = "0.8"
```
ifreund, line 19: It would nice if the drawing library remained optional. Some users of SCTK do not need decorations drawn by the library.

Cargo.toml
Darkspirit: a) rusttype could be used directly.
b) [font-kit's FAQ](https://github.com/servo/font-kit/wiki/FAQ#ok-but-i-dont-care-about-hinting-can-i-have-a-pure-rust-implementation-of-font-loading) states it would welcome a PR that adds a rusttype backend.

JMS55: Maybe you can use fontdue here instead of font-kit?

trimental:
Just a note that https://github.com/RazrFalcon/tiny-skia is something to watch for progress. It seems to be minimally dependent on other crates (hence the tiny), however the main issue is that text isn't supported. Probably need to track https://github.com/RazrFalcon/tiny-skia/issues/1. Once that issue is taken care of however it could be a very good, more powerful alternative to andrew that'll allow for more complex decorations. Another issue would be system font selection depending on whether that's implemented as well.

vberger:
Closing this PR as #183 made the underlying need outdated.

PR #134, opened 2020-06-04
chore: update rustfmt attributes

kchibisov:
It's either rust 1.44.0 broke old things entirely or removed them. Anyway, the new way is to skip formatting like that IIRC.

### DIFF
--- a/src/seat/keyboard/mod.rs
+++ b/src/seat/keyboard/mod.rs
@@ -24,8 +24,10 @@ use wayland_client::{
     Attached,
 };
 
+#[rustfmt::skip]
 mod ffi;
 mod state;
+#[rustfmt::skip]
 pub mod keysyms;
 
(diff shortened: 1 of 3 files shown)


src/seat/keyboard/mod.rs
```diff
@@ -24,8 +24,10 @@ use wayland_client::{
     Attached,
 };
 
+#[rustfmt::skip]
```
vberger, line 27: I'd prefer if those were kept in their respective modules if possible. Is a toplevel `#![rustfmt::skip]` recognized?

kchibisov: No. It applies to a following statement or something like that. So that's why they look like that.

vberger: Alright then.

PR #135, opened 2020-06-04
surface: Return `Attached<WlSurface>` instead of WlSurface

### LINKED ISSUE #132: Why does create_surface/setup_surface do surface.detach?
This causes `thread 'main' panicked at 'Attemping to create an object from a non-attached proxy.'` when trying to get a frame callback from a sctk-created surface.

I guess I should've realized that it's not `Main` but I was completely baffled by this error, wondering what the hell is going on with my frame callback for quite some time >_<

kchibisov:
We should return `Attached<WlSurface>` from `create_surface`
and `create_surface_with_scale_callback` to simplify handling
of frame callbacks and object creation, since queue could be
unavailable in a place of surface creation.

Fixes #132.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+#### Breaking Changes
+
+- `create_surface` and `create_surface_with_scale_callback` now return `Attached<WlSurface>`
+
 #### Changes
 
 - `andrew` is updated to `0.3`.
--- a/src/surface.rs
+++ b/src/surface.rs
@@ -2,7 +2,7 @@ use std::{cell::RefCell, rc::Rc, sync::Mutex};
 
 use wayland_client::{
     protocol::{wl_compositor, wl_output, wl_surface},
-    DispatchData, Main,
+    Attached, DispatchData, Main,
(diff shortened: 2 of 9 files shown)


src/surface.rs
```diff
@@ -121,7 +121,7 @@ where
         }
     });
     surface.as_ref().user_data().set_threadsafe(|| Mutex::new(SurfaceUserData::new()));
-    surface.detach()
+    (*surface).clone()
```
kchibisov, line 124: In theory such thing could be implemented in wayland-client to convert `Main<I>` to `Attached<I>`, since `Main<I>` just encapsulates `Attached<I>`.

kchibisov:
Fix for CI is #134

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 #### Changes
 
 - `andrew` is updated to `0.3`.
+- `create_surface` and `create_surface_with_scale_callback` now return `Attached<WlSurface>`
```
vberger, line 8: This is a breaking change, so it should be marked in a `#### Breaking Changes` section.

PR #136, opened 2020-06-15
Upgrade wayland-cursors to 0.26.6

Matthias-Fauconneau:
Seems to fix a panic on cursor loading on my system.
PRing this trivial change into master after hitting this for a second time ><


vberger:
This kind of update should automatically be brought by `cargo update`, since the versions are semver-compatible...

Matthias-Fauconneau:
Great! Not familiar with semver >< Sorry for the noise

PR #137, opened 2020-07-05
Advertise seats with an empty name

kchibisov:
Some compositors(plasma) do use empty name for a seat name, filtering
out by empty seat name, was leading to not working clipboard on such
compositors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 #### BugFixes
 
 - keyboard: Remove the unnecessary type parameter of `map_keyboard`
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@
 #### BugFixes
 
 - keyboard: Remove the unnecessary type parameter of `map_keyboard`
+- seat: seat not being advertise with an empty name
```
vberger, line 29:
```suggestion
- seat: seat not being advertise with an empty name
+ seat: Seats with an empty name are no longer filtered out
```

PR #138, opened 2020-07-27
event_loop: add WaylandSource::queue_mut()

kchibisov:
I actually need `queue_mut`, since I should do `sync_roundtrip` after creating a window in winit and dispatching also requires a mut EventQueue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 #### Additions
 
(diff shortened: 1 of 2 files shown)


vberger:
Ah indeed, most useful things with `EventQueue` require a `&mut` reference. I think we should rather have the `queue()` method directly return a `&mut EventQueue`, instead of adding `queue_mut()`, the current method have little reason for existing.

kchibisov:
Updated to use `mut` in `queue`.

PR #140, opened 2020-08-09
Fix fd leak in keymap updates

danieldg:
If the keymap is not passed to init_with_fd, the file descriptor is leaked.
Wrap the fd in a File earlier on so normal drop code takes care of this.

The virtual-kbd program at https://github.com/danieldg/wlroots/blob/master/examples/virtual-kbd.c triggers this leak, if you want to test.

### DIFF
--- a/src/seat/keyboard/mod.rs
+++ b/src/seat/keyboard/mod.rs
@@ -14,7 +14,12 @@
 
 #[cfg(feature = "calloop")]
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


vberger:
Oh indeed, if the keymap is locked, the file descriptor is leaked. Thanks!

Could just give a `cargo fmt` pass of your PR, so that CI turns green, before I merge it?

vberger:
Sorry for the delay, this is good now, thanks!

PR #144, opened 2020-08-18
Fix precision in keyboard repeated keys events rate

### LINKED ISSUE #141: Precision lost in repeat rate
Repeat rate gap is rounded down, which will result in incorrect delays for let's say repeat_rate 72, it will be equal to 77. To avoid that we shouldn't floor, but rather delay the conversion until we actually set a new time in ms(maybe change to micros if possible).

Related places are https://github.com/Smithay/client-toolkit/blob/dceca4d13161cee577550742a1c1e4963600a0f8/src/seat/keyboard/mod.rs#L200 and https://github.com/Smithay/client-toolkit/blob/dceca4d13161cee577550742a1c1e4963600a0f8/src/seat/keyboard/mod.rs#L480.

kchibisov:
Milliseconds are not accurate enough resulting in lower
repeat gap than it should be, however using microseconds seems enough
to solve that.

Fixes #141.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 - `Window::set_title` now truncates the provided string to 1024 bytes, to avoid blowing up
   the Wayland connection
 
+#### Bugfixes
+
+- Precision in a rate of key repeat events
+
 ## 0.10.0 -- 2020-07-10
 
 #### Breaking Changes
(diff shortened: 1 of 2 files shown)


vberger:
Thanks, I didn't expect that ms precision would not be enough tbh. ^^"

PR #146, opened 2020-08-21
Rework window.set_decorate handling

### LINKED ISSUE #143: Innability to toggle between ServerSide and None decorations.
If you try to do so, you'll get protocol error.

`zxdg_decoration_manager_v1@7: error 0: xdg_toplevel_decoration must not have a buffer at creation`.

My use case is to hide client side decorations for fullscreen windows, however since there's no way to get current decorations status, as well as hide draw hidden client side decorations, I have to do toggle between `FollowServer` and `None`, which doesn't work due to the mentioned protocol violation.

kchibisov:
This commit changes the behavior of how `window.set_decorate`
works in a following way.

If `zxdg_toplevel_decoration_v1` object is presented and alive,
requesting `None` decorations will result in setting `ClientSide`
decorations with hidden frame, and if `ClientSide` decorations were
requested, it'll result in destroying `zxdg_toplevel_decoration_v1`
object, meaning that you won't be able to get `ServerSide` decorations
back.

In case `zxdg_toplevel_decoration_v1` is not available or the
corresponding object is not alive anymore, `decorate` with `ServerSide`
or `FollowServer` values will always result in `ClientSide` decorations
being used.

Fixes #143.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,27 @@
 
 ## Unreleased
 
+#### Breaking Changes
+
+- `window.set_decorate` is now taking mutable reference
+
 #### Additions
 
-- `WaylandSource::queue` to access the `EventQueue` unserlying a `WaylandSource`
+- `WaylandSource::queue` to access the `EventQueue` underlying a `WaylandSource`
 
 #### Changes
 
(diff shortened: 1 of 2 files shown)


vberger:
Thanks! I'll handle the rebase&merge.

PR #147, opened 2020-08-21
Hide decorationos in fullscreened ConceptFrame

kchibisov:
In general there's no value in showing decorations for fullscreen
window. We also can't rely on users hidding them, since they should
manually recompute the size accounting for the decorations.

This commit also makes window to restore the size, which it had
before being fullscreened.

I've tested this on GNOME and weston with winit, and it works very good. I wonder whether we should document that concept frame isn't drawing decorations for fullscreen state?


vberger:
Thanks!

PR #148, opened 2020-08-21
Reduce the use of Mutex

kchibisov:
Certain structures are `!Send` + `!Sync` or could only
be accessed from `!Send` + `!Sync` data, meaning that
commonly used `Arc<Mutex<T>>` structure could be replaced with
`Rc<RefCell<T>>`.

I'm not entirely sure about device manager related bits, but I don't see why it should be `Arc<Mutex<_>>` for inner state.

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -1,5 +1,7 @@
 //! Window abstraction
-use std::sync::{Arc, Mutex};
+use std::cell::RefCell;
+use std::rc::Rc;
+use std::sync::Arc;
 
 use wayland_client::protocol::{
     wl_compositor, wl_output, wl_seat, wl_shm, wl_subcompositor, wl_surface,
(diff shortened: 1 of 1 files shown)


vberger:
:+1: regarding the part in `Window`

> I'm not entirely sure about device manager related bits, but I don't see why it should be Arc<Mutex<_>> for inner state.

Doesn't this allow the whole data device abstraction to be threadsafe? I tend to prefer keeping things multithread-compatible when possible.

kchibisov:
>Doesn't this allow the whole data device abstraction to be threadsafe? I tend to prefer keeping things multithread-compatible when possible.

I'll revert those parts for now, since I can't really check it precisely, but you can't do `set_selection` from the other thread, since you're creating a `Source` for that, however the rest seems to be fine...

vberger:
> but you can't do set_selection from the other thread, since you're creating a Source for that

Creating a `DataSource` requires an `Attached<WlDataDeviceManager>` yes, but nothing prevents the user from cloning the `WlDataDeviceManager` and attaching it on an other thread to create the `Source`, so I think this is fine.

PR #149, opened 2020-08-22
Add 'Option<ThemeManager>' argument to 'create_window'

kchibisov:
Allow users to pass `ThemeManager` to reduce the usage of resources
and extra loads of users' themes.

Frames could use that to detect whether they should theme cursor
over the base surface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,14 @@
 - `window.set_decorate` is now taking mutable reference
 - Added `show_window_menu` on a `Frame` trait to request a window menu for a window.
 - `ShowMenu` enum variant to `FrameRequest`
+- `create_window` now also takes `Option<ThemeManager>`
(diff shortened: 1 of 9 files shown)


vberger:
I guess the simplest approach could be to add a method to `Window()` (and thus to the `Frame` trait as well) to disable the auto-resetting of the cursor?

This way, by default the frame does what it currently does, but an app that wants precise control over its cursor can opt-out of this default behavior. What do you think about that?

kchibisov:
>This way, by default the frame does what it currently does, but an app that wants precise control over its cursor can opt-out of this default behavior. What do you think about that?

It seems like an over complication to me tbqh, since the majority of users are going to use it ThemedPointer anyway. However there's nothing wrong in adding functions to disable that behavior on a Frame I guess, since it could bring consistency to potential `Frame` writers.

kchibisov:
I've updated the PR to reflect the changes we've discussed on `#smithay`.

PR #150, opened 2020-08-23
Request window menu when right clicking on CSD decorations

kchibisov:
Some compositors like mutter can show window menu
when performing xdg_toplevel.show_mindow_menu.

That's how it looks on GNOME. (cursor is there, just wasn't captured by `grim`).

picture


vberger:
This looks good, thanks!

PR #152, opened 2020-08-27
Create subsurfaces in 'ConceptFrame' only when needed

kchibisov:
Previously 'ConceptFrame' was always creating subsurfaces, which
was consuming unnecessary resources when ServerSide decorations were
used. Now it'll create those only when frame is in fact visible.

Tested on kwin(wayland) where it fixes https://bugs.kde.org/show_bug.cgi?id=419235 , however it's still a kwin bug.
gnome(wayland)
weston
sway.

### DIFF
--- a/src/window/concept_frame.rs
+++ b/src/window/concept_frame.rs
@@ -204,7 +204,7 @@ struct PointerUserData {
  */
 
 struct Inner {
-    parts: Vec<Part>,
+    parts: Option<Vec<Part>>,
     size: (u32, u32),
     resizable: bool,
     theme_over_surface: bool,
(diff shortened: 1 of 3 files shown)


src/window/concept_frame.rs
```diff
@@ -204,7 +204,7 @@ struct PointerUserData {
  */
 
 struct Inner {
-    parts: Vec<Part>,
+    parts: Option<Vec<Part>>,
```
vberger, line 207: Maybe just keep the `Vec<_>` and test if it is empty to decide if the decorations have been initialized, rather than introducing an `Option` ?

kchibisov: I don't really mind, I've picked option since it's easier to work with and is more clear when showing that value isn't there.

vberger: true

kchibisov: Updated.

PR #153, opened 2020-08-27
Disable key repeat when getting zero rate in repeat_info

kchibisov:
Wayland spec says that zero rate in wl_keyboard.repeat_info means
that key repeat must be disabled.

Relevant spec part https://gitlab.freedesktop.org/wayland/wayland/-/blob/master/protocol/wayland.xml#L2283


vberger:
Thanks!

PR #157, opened 2020-08-29
Restore original window size after being tiled

kchibisov:
Some compositors like GNOME tile you move your window into 'hot corners'.
Once you leave a hot corner they rely on you restoring your original size.

### DIFF
--- a/src/window/mod.rs
+++ b/src/window/mod.rs
@@ -258,7 +258,15 @@ impl<F: Frame + 'static> Window<F> {
                         // Check whether we should save old size for later restoration.
                         let should_stash_size = states
                             .iter()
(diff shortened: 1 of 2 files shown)


src/window/mod.rs
```diff
@@ -258,7 +258,15 @@ impl<F: Frame + 'static> Window<F> {
                         // Check whether we should save old size for later restoration.
                         let should_stash_size = states
                             .iter()
-                            .find(|s| *s == &State::Maximized || *s == &State::Fullscreen)
+                            .find(|s| match *s {
+                                State::Maximized
+                                | State::Fullscreen
```
kchibisov, line 263: It's done that way instead of inversion of states, since the `match` is non exhaustive.

PR #159, opened 2020-09-26
Rework 'Environment' initialization methods

kchibisov:
Previously 'Environment' was initialized via 'init' method,
and then it was requiring downstream to perform two sync roundtrips
to the server, which wasn't ergonomic for rust API, however the
initialization without two roundtrips could be preferred in some
scenarios.

This commit adds new method 'Environment::new', which will take
'EventQueue' and perform proper initialization for the users. The old
method is now named 'Environment::new_pending', and you'll still have
to perform two roundtrips.

In addition, this commit renames 'init_default_environment!' to
'new_default_environment!' to respect new naming scheme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@
 #### Breaking Changes
 
 - Update `wayland-client` to version 0.28
+- `Environment::init` was renamed to `Environment::new_pending`
+- `init_default_environment!` marco was renamed to `new_default_environment!`
+
+#### Additions
+
+- `Environment::new` method to properly bootstrap environment
 
 ## 0.11.0 -- 2020-08-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@
 
 - Update `wayland-client` to version 0.28
 - `Environment::init` was renamed to `Environment::new_pending`
-- `init_default_environment!` marco was renamed to `new_default_environment!`
+- `init_default_environment!` macro was renamed to `new_default_environment!`
 
 #### Additions
 
-- `Environment::new` method to properly bootstrap environment
+- `Environment::new` method to fully bootstrap environment
 
 ## 0.11.0 -- 2020-08-30
 
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -108,8 +108,9 @@ impl<E: InnerEnv + 'static> Environment<E> {
     /// [`new_default_environment!`](../macro.new_default_environment.html) macro.
     ///
     /// `std::io::Error` could be returned if initial roundtrips to the server failed.
-    /// This call may block when doing initial roundtrips, but those blocks could only be due to
-    /// server bugs.
+    ///
+    /// If this call indefinitely blocks when doing initial roundtrips this can only be
+    /// caused by server bugs.
     pub fn new(
         display: &Attached<wl_display::WlDisplay>,
         queue: &mut EventQueue,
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@
 #### Breaking Changes
 
 - Update `wayland-client` to version 0.28
+- `Environment::init` was renamed to `Environment::new_pending`
+- `init_default_environment!` marco was renamed to `new_default_environment!`
```
vberger, line 9:
```suggestion
- `init_default_environment!` macro was renamed to `new_default_environment!`
```

CHANGELOG.md
```diff
@@ -5,6 +5,12 @@
...
+
+#### Additions
+
+- `Environment::new` method to properly bootstrap environment
```
vberger, line 13:
```suggestion
- `Environment::new` method to fully bootstrap environment
```

I'd prefer this wording, there's nothing "improper" about `new_pending`;

src/environment.rs
```diff
@@ -95,18 +97,49 @@ pub struct Environment<E> {
...
+    /// You also need to provide an instance of the inner environment type declared
+    /// using the [`environment!`](../macro.environment.html) macro.
+    ///
+    /// If you instead used the [`default_environment!`](../macro.default_environment.html), then
+    /// you need to initialize your `Environment` using the
+    /// [`new_default_environment!`](../macro.new_default_environment.html) macro.
+    ///
+    /// `std::io::Error` could be returned if initial roundtrips to the server failed.
+    /// This call may block when doing initial roundtrips, but those blocks could only be due to
+    /// server bugs.
```
vberger, line 112: *Indefinite blocking* can only be caused by server bug, a short blocking while the exchanges are done is expected.

src/environment.rs
```diff
@@ -95,18 +97,49 @@ pub struct Environment<E> {
...
     ///
     /// If you instead used the [`default_environment!`](../macro.default_environment.html), then you need
     /// to initialize your `Environment` using the
-    /// [`init_default_environment!`](../macro.init_default_environment.html) macro.
+    /// [`new_default_environment!`](../macro.new_default_environment.html) macro.
     ///
-    /// You will need to do two roundtrips of the event queue afterward to fully initialize the environment.
-    pub fn init(display: &Attached<wl_display::WlDisplay>, env: E) -> Environment<E> {
+    /// You should prefer to use `Environment::new`, unless you want to control initialization
+    /// manually or you create additional environement meaning that the initialization may be fine
```
vberger, line 138:
```suggestion
    /// manually or you create additional environment meaning that the initialization may be fine
```

vberger (changes requested):

src/lib.rs
```diff
@@ -349,7 +349,7 @@ macro_rules! init_default_environment {
             let sctk_primary_selection_manager = $crate::primary_selection::PrimarySelectionHandler::init(&mut sctk_seats);
 
             let display = $crate::reexports::client::Proxy::clone(&$display);
-            let env = $crate::environment::Environment::init(&display.attach($queue.token()), $env_name {
+            let env = $crate::environment::Environment::new_pending(&display.attach($queue.token()), $env_name {
```
vberger, line 352: I think we can replace this call by a call to `::new()` and drop the two roundtrips in the macro.

kchibisov: We need to setup primary selection manager, and since it's not a function we can't use `?` syntax here so I've done it that way to avoid manually doing `if let`.

vberger: I think I'd prefer it with the `if let`, intent is clearer that way imo. That or a `.map()` on the result, works as well.

kchibisov: I've updated PR.

vberger (approved):

PR #166, opened 2020-12-07
Replace memmap with memmap2

### LINKED ISSUE #165: RUSTSEC-2020-0077: memmap is unmaintained
After running `cargo audit` on my project, which has a transient dependence on smithay-client-toolkit, I got the following security advisory concerning the `memmap` dependency:
```
Crate:         memmap
Version:       0.7.0
Warning:       unmaintained
Title:         memmap is unmaintained
Date:          2020-12-02
ID:            RUSTSEC-2020-0077
URL:           https://rustsec.org/advisories/RUSTSEC-2020-0077
Dependency tree:
memmap 0.7.0
└── smithay-client-toolkit 0.12.0
```
From the [website](https://rustsec.org/advisories/RUSTSEC-2020-0077), either `mapr` or `memmap2` might be good replacements.

brightly-salty:
This pull request replaces the memmap dependency with the mapr crate in accordance with [this Rust security advisory](https://rustsec.org/advisories/RUSTSEC-2020-0077). I had at first proposed the memmap2 crate, but after seeing that the last release was from a year ago, switched to mapr. Closes #165.

This is my first time contributing to smithay-client-toolkit. Let me know if I need to do anything else.

Edit: Now replacing with memmap2 because of FreeBSD issues.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ bitflags = "1.0"
 nix = "0.18"
 dlib = "0.4"
 lazy_static = "1.0"
-memmap = "0.7"
+mapr = "0.8.0"
 andrew = { version = "0.3.0", optional = true }
 log = "0.4"
 wayland-client = "0.28"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ bitflags = "1.0"
 nix = "0.18"
 dlib = "0.4"
 lazy_static = "1.0"
-mapr = "0.8.0"
+memmap2 = "0.1.0"
 andrew = { version = "0.3.0", optional = true }
 log = "0.4"
 wayland-client = "0.28"
--- a/src/seat/keyboard/state.rs
+++ b/src/seat/keyboard/state.rs
@@ -1,4 +1,4 @@
-use memmap::MmapOptions;
+use mapr::MmapOptions;
 use std::{env, ffi::CString, fs::File, os::raw::c_char, os::unix::ffi::OsStringExt, ptr};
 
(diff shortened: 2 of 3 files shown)


vberger:
Hmm, it looks like `mapr` fails to compile on freeBSD. This is going to be a problem.

brightly-salty:
Looks like `mapr` uses `libc::MAP_LOCKED`, which is not supported for FreeBSD. See [docs](https://github.com/rust-lang/libc#platforms-and-documentation).

FreeBSD is not on the list of targets that `mapr` continually tests against or cross-compiles to.

Possible fixes are 
1. to open a PR on the `mapr` repository at https://github.com/danburkert/memmap-rs.
2. to try `memmap2` to see if it would support FreeBSD, or
3. keep `memmap` for now, perhaps with comments or other documentation documenting this.

brightly-salty:
@vberger Which of the three options would you prefer? I guess if this is high-priority, option 2 sounds best. It looks like the error originated from a commit specific to mapr's fork, so memmap2 should work. On the other hand, it doesn't look like it is actively maintained itself (last commit was January, and was just resetting the version to 0.1.0), which means this might just be putting the issue off until memmap2 is marked as unmaintained.

vberger:
Looks like the maintainer of memmap2 is still maintaining it, je crate just does not have many changes : https://github.com/danburkert/memmap-rs/issues/90#issuecomment-737480966 so this may be the simplest option.

brightly-salty:
@vberger Okay, it passes now. If you want me to fix the lint errors I can, but I think they're unrelated.

PR #168, opened 2020-12-19
Ready for Review: Fix font-config crash

### LINKED ISSUE #167: Crash when loading fontconfig
This has been reported to Alacritty in https://github.com/alacritty/alacritty/issues/4591.

Generally the fontconfig of sctk should probably be reworked, but even if it doesn't it should under no circumstances just prevent the entire application from working. An unwrap in sctk on something that can actually occur seems rather inappropriate.

brightly-salty:
Fixes #167 and alacritty/alacritty#4591.


src/window/concept_frame.rs
```diff
@@ -639,21 +641,40 @@ impl Frame for ConceptFrame {
...
+                                                }
                                             }
+                                            None => error!("No ttf font found"),
+                                        },
+                                        Err(err) => {
+                                            error!("No regular sans font family found: {}", err)
                                         }
+                                    },
+                                    Err(_) => {
+                                        error!("Could not load font config file")
```
chrisduerr, line 677: You can probably clean this code up quite significantly by joining the results/options together instead of handling them one by one.

I don't think it's necessary to log different messages for every single error and it gets a bit messy.

Alternatively a procedural approach is usually much easier to read too, just early returning on error and assigning the result to a variable when everything went okay. That way it doesn't get so deeply nested and difficult to match up the match arms.

chrisduerr (changes requested):

src/window/concept_frame.rs
brightly-salty: What do you think about something like this?

```
if self.font_data.is_none() {
    match get_first_ttf_regular_sans_font(font_face) {
        Some(bytes) => self.font_data = bytes,
        None => error!("No font could be found"),
      }
  }
```
```
fn get_first_ttf_regular_sans_font(font_face: String) -> Option<Vec<u8>> {
        let font_config = fontconfig::FontConfig::new().ok()?;
        let regular_family_fonts = font_config.get_regular_family_fonts(&font_face).ok()?;
        let font = regular_family_fonts
            .iter()
            .find(|p| p.extension().map(|e| e == "ttf").unwrap_or(false))?;
        std::fs::read(font).ok()
    }
```

chrisduerr: I don't really see a huge benefit over just chaining things up yourself and matching the result, but it's an option.

src/window/concept_frame.rs
```diff
@@ -639,22 +641,20 @@ impl Frame for ConceptFrame {
...
-                                    if let Ok(mut file) = ::std::fs::File::open(font) {
-                                        match file.read_to_end(&mut font_data) {
-                                            Ok(_) => self.font_data = Some(font_data),
-                                            Err(err) => {
-                                                log::error!("Could not read font file: {}", err)
-                                            }
-                                        }
-                                    }
+                                let font_bytes: Option<Vec<u8>> = fontconfig::FontConfig::new()
+                                    .and_then(|font_config: fontconfig::FontConfig| {
```
chrisduerr, line 645: Are the type annotations in these `and_then` calls necessary? I'd be surprised if they are.

brightly-salty: No, they aren't. I just included them to aid my understanding through the implementation. *Hopefully* this next commit will compile, and then it will be ready (but maybe need to squash b/c so many were failing)

brightly-salty:
I am unable to compile on my own before each commit because my build of calloop keeps failing.

brightly-salty:
@chrisduerr It's ready now for your review. It will definitely need to be squashed when it's merged because 17 commits for a couple lines is excessive.

catleeball (changes requested):
Small nit: It looks like you may have accidentally added your `.DS_Store` file. I might be good to add this to the `.gitignore` for this repository.

chrisduerr:
> I might be good to add this to the .gitignore for this repository.

Files that are specific to your development process should **always** be added to your global gitignore file, not project specific gitignores.

vberger:
@brightly-salty : if you are unable to compile `calloop`, you can disable it in SCTK by building it with
```
cargo build --no-default-features --features "frames"
```

However, the build failure of `calloop` is definitely a problem, could you report it to [calloop's repository](https://github.com/Smithay/calloop/) ?

src/window/concept_frame.rs
```diff
@@ -639,22 +640,20 @@ impl Frame for ConceptFrame {
...
+                                let fonts = font_config.and_then(|font_config| {
+                                    font_config.get_regular_family_fonts(&font_face).ok()
+                                });
+                                let font = fonts.and_then(|regular_family_fonts| {
+                                    regular_family_fonts
+                                        .iter()
+                                        .cloned()
+                                        .find(|p| p.extension().map_or(false, |e| e == "ttf"))
+                                });
+                                let font_bytes = font.and_then(|font| std::fs::read(font).ok());
```
vberger, line 653: Is this necessary to bind each intermediate result to a new name, rather than just chaining the `.and_then(...)` calls ?

src/window/concept_frame.rs
```diff
@@ -639,22 +640,20 @@ impl Frame for ConceptFrame {
...
+                                match font_bytes {
+                                    Some(bytes) => self.font_data = Some(bytes),
+                                    None => error!("No font could be found"),
```
vberger, line 656: I think we should store in `self.font_data` something indicating that the font could not be loaded if that was the case, for example by making the field an `Option<Result<Vec<u8>, ()>>`. This way, `None` indicates that the search has not been done yet, and `Some(Err(()))` indicates that it has failed.

Otherwise, SCTK will attempt again to load the font every time a redraw is triggered, which seems unnecessary.

vberger (changes requested):
Thanks for your PR!

This looks pretty good overall, There is just one main thing I left in a review comment about `self.font_data` that I think should be part of the PR as well. Once that is addressed this should be mergeable. :+1:

brightly-salty:
> @brightly-salty : if you are unable to compile `calloop`, you can disable it in SCTK by building it with
> 
> ```
> cargo build --no-default-features --features "frames"
> ```
> 
> However, the build failure of `calloop` is definitely a problem, could you report it to [calloop's repository](https://github.com/Smithay/calloop/) ?

I have reported it [here](https://github.com/Smithay/calloop/issues/31), but when I run  your build script I get similar errors, now coming client-toolkit itself.

vberger:
Oh, you're on MacOS. Well that's expected that the build fails then, as no attempt was made to have SCTK build on MacOS given Wayland is not really a thing on this platform.

.gitignore
```diff
@@ -2,3 +2,4 @@
 /target
 **/*.rs.bk
 Cargo.lock
+.DS_Store
```
vberger, line 5: As @chrisduerr noted, this kind of files should be added to [a global gitignore on your computer](https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) rather than in each project gitignore.

src/window/mod.rs
```diff
@@ -169,6 +169,7 @@ impl<F: Frame + 'static> Window<F> {
     ///
     /// Providing non `None` value for `theme_manager` should prevent theming pointer
     /// over the `surface`.
+    #[allow(clippy::match_like_matches_macro)]
```
vberger, line 172: Also please don't add unrelated changes like this. Even if the clippy builds fails, it is unrelated to your changes so you don't have to worry about it.

This should be done in an other PR, and either by fixing the code to appease the lint, or by allowing the lint at the crate level if fixing the code would require bumping the minimal rust version.

vberger (changes requested):

vberger:
Okay, this is good now, thank you!

catleeball:
Thank you for making a PR to address my [alacritty issue](https://github.com/alacritty/alacritty/issues/4591)! :tada:

chrisduerr:
@vberger Do you want to release a patch for this or are there any other things in the pipeline?

vberger:
You're right, there nothing else waiting currently. `0.12.2` is now released.

chrisduerr:
Awesome, thanks for the quick response!

PR #172, opened 2020-12-28
Remove byteorder dependency

psychon:
This PR removes the dependency on the byteorder crate.

My motivation is a bad one: Less dependencies! (and byteorder is not really necessary for the things that are done here)
My testing is weak: `cargo build --all-targets && cargo test` succeeds. I did not even run any of the examples since I do not have a wayland compositor at hand.

I did this in separate commits so that in case something is wrong, `git bisect` is more helpful. However, I am not sure if that is really necessary. If wanted, I can squash this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+#### Changes
+
+- Dependency on `byteorder` was replaced with `u32::from_ne_bytes()`
+
 ## 0.12.1 -- 2020-12-08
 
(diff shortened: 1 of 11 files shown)


vberger:
Right, the stdlib allows us to drop this dependency now. Thanks!

vberger:
Could you just add a changelog entry notifying that `byteorder` is no longer a dependency?

psychon:
Done

vberger:
Perfect, thanks!

PR #173, opened 2021-01-19
Set F_SEAL_SHRINK on memfds we create

danieldg:
This prevents us from ever shrinking the file (which we never do) and lets the
wayland server skip configuring its SIGBUS handler when accessing our buffers.


vberger:
Sounds good on principle, but does the server have a way to actually check if the FD is sealed before accessing it? I'm not very familiar with these APIs.

danieldg:
Yes, you can read them with `F_GET_SEALS` - grep for `sigbus_is_impossible` in the wayland C code for what this allows.

vberger:
I see, thanks!

PR #174, opened 2021-01-27
Implement `start_interactive_move`.

daxpedda:
As discussed in chat.

I have copied the documentation from the wayland-protocols specification, as I have basically very little clue about Wayland I'm open to suggestions, thank you.

~~I will leave it as a draft until I try it out, which I'm in the process of doing at the moment.~~
Tested with https://github.com/rust-windowing/winit/pull/1840, works like a charm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+#### Additions
+
+- `Window::start_interactive_move` to enable dragging the window with a user action
+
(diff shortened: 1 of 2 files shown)


vberger:
Thanks, that looks good!

I think we can remove the last paragraph of the doc-comment, which does not add much for users of SCTK. And could you please also add an entry in the "unreleased" section of the changelog for this?

vberger:
Thanks!

PR #176, opened 2021-01-31
Add optional xdg_output support

danieldg:
This adds name and description fields to OutputInfo and populates them if a properly initialized XdgOutputHandler is present.  If the global is not present or not offered by the server, output handling is unchanged and the new fields will be blank.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- Batch output information updates instead of potentially making multiple
+  callbacks for one logical change
+
 #### Additions
 
 - `Window::start_interactive_move` to enable dragging the window with a user action
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 - Batch output information updates instead of potentially making multiple
   callbacks for one logical change
 
+#### Breaking Changes
+
+- Mark OutputInfo as `#[non_exhaustive]` to allow future expansion without
+  breaking API.
+
 #### Additions
 
 - `Window::start_interactive_move` to enable dragging the window with a user action
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Batch output information updates instead of potentially making multiple
   callbacks for one logical change
+- Add name and description fields to OutputInfo.
 
 #### Breaking Changes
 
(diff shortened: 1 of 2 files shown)


vberger:
Thanks, this looks good overall, I just want to make sure I understand correctly the intended behavior: what is expected here seems to be that the server will send the `wl_output`-related events, then `wl_output.done`, then the `xdg_output` related events, then an other `wl_output.done`, is that correct? Is this behavior guaranteed by the protocol?

Also, this needs to be introduced in the changelog, and I think it is a breaking change, since some public fields are added and the struct is make `#[non-exhaustive]`.

danieldg:
> Thanks, this looks good overall, I just want to make sure I understand correctly the intended behavior: what is expected here seems to be that the server will send the `wl_output`-related events, then `wl_output.done`, then the `xdg_output` related events, then an other `wl_output.done`, is that correct? Is this behavior guaranteed by the protocol?

This is what I observed in testing.  It is also the only possible result if the incoming messages are processed independently; the binding of an wl_output is required to produce wl_output events (followed by Done), and while processing that binding there is no indication that xdg_output events need to be sent.

However, I don't see any statements that prohibit a compositor from combining the incoming events before processing them and then combining the output events (and eliminating the second wl_output.done).  I'm not sure this type of queue reordering is allowed by wayland; it would be very confusing if it were because it would produce different event streams depending on how the OS splits up send and recv calls on the unix socket.

If this ever happened, the unreachable!() case would be hit in process_xdg_event, since the Name event is mandatory for xdg_output.  I guess that case could be expanded to convert OutputData::Pending to OutputData::PendingXDG so that the one Done produces the notification.

> Also, this needs to be introduced in the changelog, and I think it is a breaking change, since some public fields are added and the struct is make `#[non-exhaustive]`.

It is; that's one reason why I split out that commit, so that it could be included earlier if needed.  Should I mark this as a 0.13 change in the changelog?  I could also add #[non_exhaustive] to Mode, though I can't think of any reason for adding fields to that struct - but I think it's the only other one that doesn't already have a private field.

vberger:
Just introduce a changelog entry in the "unreleased", creating a new **Breaking changes** subsection (see for example release 0.12) indicating what was added and why it is a breaking change. The rest sounds good, thank ! :+1:

vberger:
Thanks!

PR #178, opened 2021-02-14
Allow linking to libxkbcommon at compile time

milkey-mouse:
I am packaging a version of [alacritty](https://github.com/alacritty/alacritty) that runs on Wayland (without Xwayland) for [Guix](https://guix.gnu.org/), a distro which, like NixOS, stores all shared libraries in a [weird place](https://guix.gnu.org/manual/en/html_node/The-Store.html). The dynamic linker doesn't check this path by default, so sometimes packagers are forced to modify the library search path after building, but a better way is to link to the library at compile time.

This commit introduces a "dlopen" feature that is enabled by default that preserves the existing behavior of not linking to libxkbcommon except for trying to dlopen it at runtime. If the feature is disabled, cargo will link to the system's libxkbcommon.so at compile time.

To do this, I used the [generic macros in dlib](https://github.com/vberger/dlib/blob/7352a7731b687dcba6446c8c8362a7c13f409fcb/README.md#remaining-generic-in-your-crate) which either dlopen the library in a lazy_static struct like the current behavior, or invoke the "statically dynamically-linked" functions if the dlopen feature is turned off.

### DIFF
--- a/src/seat/keyboard/ffi.rs
+++ b/src/seat/keyboard/ffi.rs
@@ -174,7 +174,7 @@ bitflags::bitflags!(
     }
 );
 
-dlopen_external_library!(XkbCommon,
+external_library!(XkbCommon, "xkbcommon",
 functions:
     fn xkb_keysym_get_name(xkb_keysym_t, *mut c_char, usize) -> c_int,
     fn xkb_keysym_from_name(*const c_char, xkb_keysym_flags) -> xkb_keysym_t,
@@ -253,8 +253,9 @@ functions:
     fn xkb_compose_state_get_one_sym(*mut xkb_compose_state) -> xkb_keysym_t,
 );
 
+#[cfg(feature = "dlopen")]
 lazy_static::lazy_static!(
-    pub static ref XKBCOMMON_OPTION: Option<XkbCommon> = {
+    pub static ref XKBCOMMON_OPTION: Option<XkbCommon> = unsafe {
         XkbCommon::open("libxkbcommon.so.0")
             .or_else(|_| XkbCommon::open("libxkbcommon.so"))
             .ok()
--- a/src/seat/keyboard/state.rs
+++ b/src/seat/keyboard/state.rs
@@ -1,7 +1,11 @@
 use memmap2::MmapOptions;
 use std::{env, ffi::CString, fs::File, os::raw::c_char, os::unix::ffi::OsStringExt, ptr};
 
-use super::ffi::{self, xkb_state_component, XKBCOMMON_HANDLE as XKBH};
+#[cfg(feature = "dlopen")]
+use super::ffi::XKBCOMMON_HANDLE as XKBH;
+#[cfg(not(feature = "dlopen"))]
+use super::ffi::*;
+use super::ffi::{self, xkb_state_component};
 use super::Error;
 
 pub(crate) struct KbState {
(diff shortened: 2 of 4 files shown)


milkey-mouse:
Sorry about the force-pushes. CI was having some trouble and I haven't used Ubuntu in a while.

vberger:
Thanks for the PR, and sorry for the delay.

If my memory serves me well, I initially made xkbcommon dlopen-only because I was worried about weird interactions with wayland-client, which uses `dlib` behind the hood as well. It is entierely possible that this is no longer a concern, given the numerous refactors that occured since.

To make everything clean, I think first, the `dlopen` feature of SCTK should imply `wayland-client/dlopen` as well. But also, we should ensure that everything goes well if a project depends on both SCTK and wayland-client, but ony activates the `dlopen` feature of wayland-client. Could you check this?

milkey-mouse:
> To make everything clean, I think first, the `dlopen` feature of SCTK should imply `wayland-client/dlopen` as well.

Done.

> But also, we should ensure that everything goes well if a project depends on both SCTK and wayland-client, but ony activates the `dlopen` feature of wayland-client.

I think this is actually impossible without changing dlib. As you probably know Cargo features are additive and it'll only build one copy of dlib with the union of all features enabled in dependent crates. So if `wayland-client/dlopen` is enabled, `dlib/dlopen` is enabled, but `smithay-client-toolkit` has no way of knowing whether that feature is enabled by some other package.

I do have an idea how to fix this, but it'd require changes to dlib: in the next major version, remove the `dlopen` feature from dlib entirely and move `#[cfg(feature = "dlopen")]` *inside* the generic `external_library!` macro, something like this:

```rust
#[macro_export]
macro_rules! ffi_dispatch(
    ($handle: ident, $func: ident, $($arg: expr),*) => (
        {
            // TODO: #[cfg] on statements will only work on rust 1.42 and up:
            // https://github.com/rust-lang/rust/pull/69201/
            #[cfg(feature = "dlopen")]
            let ret = ($handle.$func)($($arg),*);
            #[cfg(not(feature = "dlopen"))]
            let ret = $func($($arg),*);

            ret
        }
    )
);

#[macro_export]
macro_rules! ffi_dispatch_static(
    ($handle: ident, $name: ident) => (
        {
            #[cfg(feature = "dlopen")]
            let ret = $handle.$name;
            #[cfg(not(feature = "dlopen"))]
            let ret = &$name;

            ret
        }
    )
);

#[macro_export]
macro_rules! external_library(
    ($structname: ident, $link: expr,
        $(statics: $($sname: ident: $stype: ty),+,)|*
        $(functions: $(fn $fname: ident($($farg: ty),*) -> $fret:ty),+,)|*
        $(varargs: $(fn $vname: ident($($vargs: ty),+) -> $vret: ty),+,)|*
    ) => (
        #[cfg(feature = "dlopen")]
        dlopen_external_library!(
            $structname, $(statics: $($sname: $stype),+,)|*
            $(functions: $(fn $fname($($farg),*) -> $fret),+,)|*
            $(varargs: $(fn $vname($($vargs),+) -> $vret),+,)|*
        );

        #[cfg(not(feature = "dlopen"))]
        link_external_library!(
            $link, $(statics: $($sname: $stype),+,)|*
            $(functions: $(fn $fname($($farg),*) -> $fret),+,)|*
            $(varargs: $(fn $vname($($vargs),+) -> $vret),+,)|*
        );
    );
);
```

This would require each package depending on dlib to have its own `dlopen` feature, which might be too much of a hack. I'm not sure how else to make this possible though.

vberger:
Actually, this is an interesting idea. The `#[cfg(...)]` statements inside the macro resolve using the cargo features of the crate the macro was expanded into, I didn't realize that. That would allow fine-control over which libraries are dlopen-ed, and which are linked.

milkey-mouse:
OK, I'll make a PR for dlib. Would you be OK with increasing the minimum supported Rust version to 1.43 for the next dlib release? It would be nice not to have to work around the [lack of support](https://github.com/rust-lang/rust/pull/69201/) for `#[cfg]` directly on statements.

milkey-mouse:
> It would even be possible to allow customization of the feature name, to let the crate control everything. Much better than relying on the good propagation of cargo features.

I was about to add this, but I realized you can already do this easily in a dependent crate (in fact, this was how I was originally going to fix the SCTK/wayland-client dlopen problem before I thought of changing the macro) like this:

```rust
// in your own crate, depending on dlib:
#[cfg(feature = "dlopen-mylib")]
dlopen_external_library!(MyLib);
#[cfg(not(feature = "dlopen-mylib"))]
link_external_library!("mylib");
// you no longer care whether dlib's dlopen feature is enabled
```

src/seat/keyboard/state.rs
```diff
@@ -210,18 +245,21 @@ impl KbState {
...
-            None => return Err(Error::XKBNotFound),
+        #[cfg(feature = "dlopen")]
+        {
+            if ffi::XKBCOMMON_OPTION.as_ref().is_none() {
+                return Err(Error::XKBNotFound);
+            }
+        }
+        let context = unsafe {
+            ffi_dispatch!(XKBH, xkb_context_new, ffi::xkb_context_flags::XKB_CONTEXT_NO_FLAGS)
         };
```
vberger, line 256: Was there a particular need for renaming this variable?

milkey-mouse: Yes. Otherwise it shadows the `keyboard::ffi::xkb_context` function when the `dlopen` feature is disabled. I suppose this could also be fixed by a more careful use statement but it would get unwieldy considering the large number of functions from the ffi module used in this file. I was just following the instructions in the dlib readme that said to do something like this:

```rust
#[cfg(feature = "dlopen-foo")]
use ffi::FOO_STATIC;
#[cfg(not(feature = "dlopen-foo"))]
use ffi::*;
```

src/seat/keyboard/ffi.rs
```diff
@@ -174,7 +174,8 @@ bitflags::bitflags!(
     }
 );
 
-dlopen_external_library!(XkbCommon,
+#[cfg_attr(not(features = "dlopen"), link(name = "xkbcommon"))]
+external_library!(XkbCommon, "xkbcommon",
```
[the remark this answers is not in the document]
milkey-mouse: this should have already been removed, looks like it slipped back in during amending

src/seat/keyboard/state.rs
vberger: Oh, ok, alright.

vberger:
Looks good, thanks!

PR #180, opened 2021-02-26
Use HTTPS

atouchet:


src/seat/keyboard/keysyms.rs
```diff
@@ -1315,7 +1315,7 @@ pub const XKB_KEY_Greek_switch                  :u32 = 0xff7e;  /* Alias for mod
 
 /*
  * Technical
- * (from the DEC VT330/VT420 Technical Character Set, http://vt100.net/charsets/technical.html)
+ * (from the DEC VT330/VT420 Technical Character Set, https://vt100.net/charsets/technical.html)
```
vberger, line 1318: This file is generated by a script from system headers, there is no point in modifying it.

vberger:
Thanks.

PR #181, opened 2021-03-04
use font-loader instead of andrew's fontconfig

### LINKED ISSUE #170: Title fails to show in a flatpak
I'm using sctk via winit, and when I run my program (https://github.com/JMS55/sandbox) as a flatpak, the title text doesn't render. 
image

vberger:
This changes SCTK's logic to use [font-loader](https://crates.io/crates/font-loader) instead of the in-house fontconfig logic of andrew to find & load fonts. font-loader uses the fontconfig library, so this should hopefully fix most font loading issues.

This is a hotfix measure in waiting for something like #131 to move forward.

Should fix #170


daxpedda:
What was the reason `font-config` was chosen over `font-loader`?

vberger:
fontconfig is a lower-level dependency (font-loader depends on it), and font-loader appears to be mostly unmaintained now.

daxpedda:
I missed that somehow, sorry.
What do you think of [rust-fontconfig](https://github.com/fschutt/rust-fontconfig)? It's still very new though.

vberger:
I was not aware of it, will have to give it a look. Thanks for the heads-up

PR #184, opened 2021-03-14
Add an alternate AutoMemPool for allocating buffers

danieldg:
This tracks the allocation of regions in a MemPool along with the
Wayland server's release operations instead of requiring a callback to
notify the client that reuse is possible.

Note that even with the normal MemPool, clients should prefer using
wl_surface::frame to delay draw operations instead of using the SHM
release callback.


vberger (approved):
Apart from a missing changelog entry, this PR is good as is. It seems to behave well under the few tests I gave it.

vberger:
Thanks!

PR #189, opened 2021-04-30
Upgrade calloop version to `0.7.2`

yilinwei:
Apologies for the formatting problems @vberger. I've got an auto-formatter running on the files and can't work out the correct `.rustfmt.toml` for the current code; do you mind sharing the settings please?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   `FallbackFrame`. Dependency on `andrew` and `fontconfig` is dropped in the process. If fancier
   decorations are needed, they should be implemented using the `Frame` trait.
 
+- `calloop::Source` is switched to `calloop::RegistrationToken` as a 1-1 replacement in all public interfaces.
+
 #### Additions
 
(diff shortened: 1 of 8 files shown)


vberger:
Hi, thanks for the PR!

The formatting config is that of the `rustfmt.toml` file that this committed on the repo, could you please setup your formatter to follow it?

yilinwei:
@vberger 

Major :facepalm:. I have no idea how my is setup ignored it at the beginning, and I assumed that it didn't exist so I copy-pasted a `.rustfmt.toml` which apparently takes precedence from the command line. Thanks for your patience.

vberger:
Thanks, that looks good! Have you tested that the examples that use the calloop integration still work as expected, as a sanity-check?

Also, could you please add a changelog entry about this change (which is a breaking change)?

yilinwei:
I've tested each of the examples which had a `calloop` dependency as a sanity check with sway on my particular OS. I have not done any other extensive tests.

I've added an entry in `CHANGELOG.md` as well.

vberger:
Thanks!

PR #190, opened 2021-05-29
Add minimum alignment guarantees to buffers from AutoMemPool

danieldg:
This mostly already happens, but it is useful to guarantee it, especially if rendering directly to the shm canvas.


vberger:
Sorry for the (very) late review. This is good, thanks a lot!

PR #193, opened 2021-06-30
Update calloop to v0.9.0

cmeissl:
This PR updates calloop to v0.9.0 and fixes the remaining warnings and clippy errors

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -67,52 +67,55 @@ impl EventSource for WaylandSource {
 
     fn process_events<F>(
         &mut self,
-        _: calloop::Readiness,
-        _: calloop::Token,
+        readiness: calloop::Readiness,
+        token: calloop::Token,
         mut callback: F,
     ) -> std::io::Result<PostAction>
     where
         F: FnMut((), &mut EventQueue) -> std::io::Result<u32>,
     {
-        // in case of readiness of the wayland socket we do the following in a loop, until nothing
-        // more can be read:
-        loop {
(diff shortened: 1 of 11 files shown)


src/event_loop.rs
```diff
@@ -111,19 +112,23 @@ impl EventSource for WaylandSource {
             // quickly. Either it is slowed down or we are a spammer.
             // Should not really happen, if it does we do nothing and will flush again later
         }
-        Ok(())
+        Ok(PostAction::Continue)
```
vberger, line 115: Due to the new semantics of calloop, for full correctness the whole body of this function needs to be wrapped En the callback of `self.fd.process_events()`.

cmeissl: Ok, moved into `self.fd.process_events`.

cmeissl: Are you ok with the changes or is there anything else that needs to be changed?

vberger:
Thanks! However it looks like some imports are breaking the CI. I think you need to add a few `#[cfg()]` attributes so that they are only imported when the feature is enabled.

l4l:
`calloop` updates nix package to 0.21 ([link](https://github.com/Smithay/calloop/blob/master/Cargo.toml#L19)), would be handy to update there as well.

vberger:
Actually there is also `0.22` pending as well.

cmeissl:
> Thanks! However it looks like some imports are breaking the CI. I think you need to add a few `#[cfg()]` attributes so that they are only imported when the feature is enabled.

Should be fixed now.

PR #198, opened 2021-08-23
use pkg-config to properly link to xkbcommon library when build without dlopen feature

semarie:
some OS (like OpenBSD or NetBSD) places third-parties libraries in no default paths, and requires explicit configuration to link to them. It makes smithay-client-toolkit difficult to build on such systems, as when not using `dlopen` feature, the build will fail.

For example, with alacritty build:
```
Building [=======================> ] 188/189: alacritty(bin)

error: linking with `cc` failed: exit status: 1
  |
  = note: [annoying stuff]
  = note: ld: error: unable to find library -lxkbcommon
          cc: error: linker command failed with exit code 1 (use -v to see invocation)

error: aborting due to previous error

error: could not compile `alacritty`

To learn more, run the command again with --verbose.
```

Currently, there is no way to specify a directory in smithay-client-toolkit to found the library and properly link to it.

This PR uses `pkg-config` crate to configure build script to properly locate the library based on pkg-config configuration file.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -2,7 +2,5 @@ extern crate pkg_config;
 
 fn main() {
     #[cfg(not(feature = "dlopen"))]
-    if pkg_config::Config::new().find("xkbcommon").is_ok() {
-        return;
-    }
+    pkg_config::Config::new().find("xkbcommon").unwrap();
 }
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,8 @@
+extern crate pkg_config;
+
+fn main() {
+    #[cfg(not(feature = "dlopen"))]
+    if pkg_config::Config::new().find("xkbcommon").is_ok() {
+        return;
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+#### Bugfixes
+
+- when not using `dlopen` feature, `xkbdcommon` library is linked using `pkg-config`
+
 ## 0.15.0 - 2021-08-10
 
 #### Breaking Changes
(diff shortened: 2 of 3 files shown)


semarie:
for reference, the problem was originally reported to https://github.com/alacritty/alacritty/issues/5422

build.rs
```diff
@@ -0,0 +1,8 @@
+extern crate pkg_config;
+
+fn main() {
+    #[cfg(not(feature = "dlopen"))]
+    if pkg_config::Config::new().find("xkbcommon").is_ok() {
+        return;
+    }
```
vberger, line 7: What is the point of this `if` block? Shouldn't the build script panic if `xkbcommon` cannot be found instead?

vberger (changes requested):
Thanks a lot, this is a clear oversight of when we added support for directly linking to xkbcommon rather than always dlopen-ing it.

I just have a small issue with the build-script logic, and could you also add an entry in the changelog, under the `## Unreleased` section and as a `### Bugfixes` subection (following the previous entries structure) ?

build.rs
semarie: I tried to avoid to break existing and working behaviour: just calling `.unwrap()` would panic the build if `xkbcommon.pc ` (the configuration file used by pkg-config) isn't found but `libxkbcommon.so` exists.

But I am unaware if such configuration exists or if it is just an ill-case.

semarie: after looking at xkbcommon repository, it seems pkgconfig is generated since at least 4 years (introduction of meson.build). so it should be an ill-case.

I will use `unwrap()`, and it should also solve the problem of build with 1.42 (where conditionnal block on conditionnal isn't supported).

semarie:
feel free to told me if you present I push amended commit instead of adding new commits on top of previous one.

vberger (approved):
Okay, this is good for me, thanks!

Regarding the commits, I tend to try maintaining a succinct git history, so if you can amend your PR to make it as a single commit that would be appreciated. Otherwise I'll just squash-merge it.

PR #200, opened 2021-08-26
Implement fmt::Debug nearly everywhere

i509VCB:
There is one spot that could be inlined to a derive, but this requires wayland-cursor getting some `Debug` also.

This also sets the `missing_debug_implementations` lint to warn


vberger:
Thanks!

PR #203, opened 2021-10-11
WLR layer surfaces

Eskpil:
I added a very tiny abstraction for creating wlr shells. It does nothing special, it just eliminates 50-100 lines of code when creating a wlr shell.

### DIFF
--- a/src/shell/layer.rs
+++ b/src/shell/layer.rs
@@ -1,4 +1,4 @@
-/// Wlr shell abstraction
+/// WLR layer abstraction
 use std::cell::Cell;
 use std::rc::Rc;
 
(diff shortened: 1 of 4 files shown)


src/shell/wlr.rs
```diff
@@ -0,0 +1,154 @@
...
+            Layer::Bottom => zwlr_layer_shell_v1::Layer::Bottom,
+            Layer::Top => zwlr_layer_shell_v1::Layer::Top,
+            Layer::Overlay => zwlr_layer_shell_v1::Layer::Overlay,
+        }
+    }
+}
+
+/// A struct representing the wlr shell
+
+pub struct WlrShell {
```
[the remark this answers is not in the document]
Eskpil: Yeah i agree. A layer surface makes way more sense.

src/shell/wlr.rs
```diff
@@ -0,0 +1,154 @@
...
+    Top,
+
+    /// All the way in the front, overiding what other applications are trying to draw.
+    Overlay,
+}
+
+/// Just a tiny abstraction making anchors easier. Anchor says something about where on the screen the shell is rendered
+
+#[derive(PartialEq, Copy, Clone)]
+pub enum Anchor {
```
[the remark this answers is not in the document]
Eskpil: I think i will just reexport them and dont bother with it.

src/shell/layer.rs
```diff
@@ -0,0 +1,110 @@
...
+};
+
+pub use wayland_protocols::wlr::unstable::layer_shell::v1::client::{
+    zwlr_layer_shell_v1::Layer, zwlr_layer_surface_v1::Anchor,
+};
+
+/// Render event
+
+#[derive(PartialEq, Copy, Clone)]
+pub enum RenderEvent {
```
i509VCB, line 21: Implement `Debug`?

src/shell/layer.rs
```diff
@@ -0,0 +1,110 @@
...
+        height: u32,
+    },
+
+    /// Surface has closed and wants to be closed here also.
+    Closed,
+}
+
+/// A struct representing the layer surface (wlr)
+
+pub struct LayerSurface {
```
i509VCB, line 37: `Debug` here also

src/shell/layer.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        /// The new height of the surface
+        height: u32,
+    },
+
+    /// Surface has closed and wants to be closed here also.
+    Closed,
+}
+
+/// A struct representing the layer surface (wlr)
```
i509VCB, line 35: I think it would be useful to go into detail about what I layer is in the context of Wayland and possible use cases for a layer

src/shell/layer.rs
```diff
@@ -0,0 +1,110 @@
...
+
+pub struct LayerSurface {
+    /// The raw wl_surface
+    pub surface: wl_surface::WlSurface,
```
i509VCB, line 39: I would probably make the surface, layer, anchor and dimensions accessible via functions similar to how the `Window` abstraction is. The `surface()` function specifically returning an `&wl_surface::WlSurface`

src/shell/layer.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    pub(crate) layer_surface: Main<zwlr_layer_surface_v1::ZwlrLayerSurfaceV1>,
```
i509VCB, line 41: This is fine for now, we can always introduce a `LayerInner` when ext-layer-shell is initiated into wayland-protocols

src/shell/layer.rs
```diff
@@ -0,0 +1,110 @@
...
+        layer_shell: &Attached<zwlr_layer_shell_v1::ZwlrLayerShellV1>,
+        layer: Layer,
+        anchor: Anchor,
+        dimensions: (u32, u32),
+    ) -> Self {
+        let layer_surface =
+            layer_shell.get_layer_surface(&surface, Some(output), layer, "example".to_owned());
+
+        layer_surface.set_size(dimensions.0, dimensions.1);
+        // Anchor to the top left corner of the output
```
i509VCB, line 71: This comment probably doesn't apply

src/shell/layer.rs
```diff
@@ -0,0 +1,110 @@
...
+                    if next != Some(RenderEvent::Closed) =>
+                {
+                    layer_surface.ack_configure(serial);
+                    next_render_event_handle.set(Some(RenderEvent::Configure { width, height }));
+                }
+                (_, _) => {}
+            }
+        });
+
+        // Commit so that the server will send a configure event
```
i509VCB, line 91: Initial commit for configure I guess?

src/shell/layer.rs
```diff
@@ -0,0 +1,110 @@
+/// WLR layer abstraction
```
i509VCB, line 1: Note that module doc comments are `//!`. You can always run `cargo doc` to check if the docs appear in the right spot.

This would also be a great place to summarize the scope of the layer module and provide a code example. The code example is probably best derived from the examples you've written.

src/shell/layer.rs
i509VCB, line 21: I'd probably go for `LayerEvent` since `Close` isn't exactly an event related to rendering.

src/shell/layer.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub layer: Layer,
+
+    /// Where on the screen it is positioned
+    pub anchor: Anchor,
+
+    /// The dimensions of the wlr surface
+    pub dimensions: (u32, u32),
+
+    /// The next render event
+    pub render_event: Rc<Cell<Option<RenderEvent>>>,
```
i509VCB, line 53: I would mimic what `Window` does here with callbacks.

Using a callback here makes sense and also provides the `DispatchData` the client has provided.

This would mean your LayerSurface::new would also now need to take a `FnMut(Event, DispatchData) + 'static` parameter.

Eskpil: Could you elaborate?

i509VCB: If you look at the methods to create a window you will see one of the parameters is a `Impl` which has type bounds defined in the where statement.

This FnMut the user passes in is essentially a lambda that is invoked when some event occurs on the window.

You'll see where you call quick_assign on the layer you create that you are given an event, that is where you will invoke the callback.

The DispatchData is a dynamic data container for some arbitrary data a user of the library passes in that is available when events come in.

src/shell/wlr.rs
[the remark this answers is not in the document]
i509VCB: Given the wlr layer shell is likely to be stabilized in the future, I would actually declare this bitflag/enum in module and just implement `From` for the wlr and ext and the sctk type so we don't need to break clients when the ext extension is initiated.

vberger:
@Eskpil did you by chance forget to push a commit? There are a few comments you marked as resolved but that are not addressed in the current state of the PR.

PolyMeilex:
Layer shell has been supported for a while now: https://docs.rs/smithay-client-toolkit/latest/smithay_client_toolkit/shell/wlr_layer/index.html therefore I'm closing this. Either way, thank you for your work, and have a nice day!

PR #206, opened 2021-10-15
Very minor documentation improvement

WhyNotHugo:
This tiny detail isn't mentioned and can possible save a lot of headaches.

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -15,7 +15,7 @@ use wayland_client::EventQueue;
 /// will be given access to the `EventQueue` and you should call `.dispatch_pending()`
 /// and forward its return value, allowing you to handle orphan events as you prefer.
 ///
-/// If you don't use orphan events, the `quick_insert` method will firectly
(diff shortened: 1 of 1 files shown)


src/event_loop.rs
```diff
@@ -15,9 +15,12 @@ use wayland_client::EventQueue;
 /// will be given access to the `EventQueue` and you should call `.dispatch_pending()`
 /// and forward its return value, allowing you to handle orphan events as you prefer.
 ///
-/// If you don't use orphan events, the `quick_insert` method will firectly
+/// If you don't use orphan events, the `quick_insert` method will directly
 /// insert the source into a provided `LoopHandle` with an adapter which will panic
 /// whenever an oprhan event is encountered.
+///
+/// The event loop's global shared data will be passed as `dispatch_data` too all
+/// the event callbacks.
```
vberger, line 23: Thanks, however this should be added on the `quick_insert()` method as it is only true when using it.

PR #209, opened 2021-12-08
Dumb refactor : SimpleGlobal --> SingleGlobal && GlobalHandler --> SingleGlobalHandler

Maxzor:
Hello,
I don't know if this is welcome, but this helps me a bit in my discovery of the code... it really looked like a dichotomy between single and multi.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,13 @@ lazy_static = "1.0"
 memmap2 = "0.3.0"
 log = "0.4"
 wayland-client = "0.29"
-wayland-protocols = { version = "0.29" , features = ["client", "unstable_protocols"] }
+wayland-protocols = { version = "0.29", features = [
+    "client",
+    "unstable_protocols",
+] }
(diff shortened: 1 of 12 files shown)


i509VCB:
If @vberger wants to ensure existing users don't break, we could probably get away with advertising the old names under some typedef with a deprecated attribute so code doesn't break?

Do wait for his response before doing that, I may be entirely wrong.

Maxzor:
> If @vberger wants to ensure existing users don't break [...]

Ah yes, first commit, already breaking userspace :man_facepalming:

vberger:
Hi, thanks for your PR.

I understand where you're coming from with this, the names of these types are not super clear. However, as @i509VCB noted those are breaking change, and the timing is not ideal: we are working on significant API rework of SCTK as a followup of https://github.com/Smithay/wayland-rs/pull/381, and those types are going away in the process. As such, I'm not sure it's worth doing a release renaming them right now.

Issue #186, opened 2021-03-23
SimpleGlobal shouldn't blindly bind the latest version of the interface


tadeokondrak:
Changes to Wayland interfaces aren't forward compatible, and there have been multiple cases of a Wayland protocol changing behaviour between versions (grep SINCE_VERSION in wlroots for examples.)

`SimpleGlobal` doesn't provide a way to limit what version it binds, so clients can break when the compositor upgrades a protocol. It should probably have a "maximum version" parameter at least, or maybe a version range.

danieldg:
Note: the Interface traits already have a VERSION associated const corresponding to the protocol version they were compiled from.  This should also be used as a maximum - so clients *may* request a lower version than VERSION, but it should be an error to request a higher one, and the actual version used should be min(server's version, protocol XML version, client's requested version).

PR #211, opened 2021-12-24
Add max_version parameter to SimpleGlobal

tadeokondrak:
This is needed because new versions of protocols can change in a
backwards-incompatible manner, and clients unaware of the changes can
crash or behave incorrectly.

Note that Interface::VERSION on its own is not enough for this purpose,
because a protocol can make a backwards-incompatible change affecting
only the specified behaviour of the client without changing the interface
description.

Fixes https://github.com/Smithay/client-toolkit/issues/186

### DIFF
--- a/examples/layer_shell.rs
+++ b/examples/layer_shell.rs
@@ -134,7 +134,7 @@ impl Drop for Surface {
 
 fn main() {
     let (env, display, queue) =
-        new_default_environment!(Env, fields = [layer_shell: SimpleGlobal::new(),])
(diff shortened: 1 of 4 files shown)


vberger:
Thanks for you PR! This is indeed the correct fix for this long-standing issue.

However, right now the timing isn't ideal as it is a breaking change. We are working on deep changes to SCTK as a follow-up of https://github.com/Smithay/wayland-rs/pull/381 and this is going to be fixed as part of the deep API rework, so I'm not sure it is work making a small breaking release with this change before that.

Issue #212, opened 2021-12-25
Huge slowdowns with wayland on nvidia gpus


Molytho:
The new NVidia driver introduced a huge slowdown in applications using client-toolkit.
As mentioned here https://github.com/NVIDIA/egl-wayland/issues/47#issuecomment-977184138 the issue is this part of the code:
https://github.com/Smithay/client-toolkit/blob/817ebb4c74dee7985411820e9c90fd09ea3e1027/src/event_loop.rs#L87.

The WaylandSource code has two issues:

1. https://github.com/Smithay/client-toolkit/blob/817ebb4c74dee7985411820e9c90fd09ea3e1027/src/event_loop.rs#L89
might block resulting in starvation or huge slowdowns of the calloop loop.
This is trivially fixable with something like [this](https://github.com/Molytho/client-toolkit/commit/85a2ee1b682be2560efa7c4afdc030f686e599ac).
2. This is the huge issue:
We (very often) end up blocking in calloop waiting for events on the display file descriptor. This happens when we finished dispatching while another thread reads new data from the file descriptor. There is a fix for this: When we intend to block on the display file descriptor we are expected to grab a read lock with https://github.com/Smithay/client-toolkit/blob/817ebb4c74dee7985411820e9c90fd09ea3e1027/src/event_loop.rs#L87

So, to fix this we need a way to run code *before* poll is called by calloop which would fix both issues and thus make the mentioned workaround of 1 useless. Also we need a way to release the lock when no event were ready on the file descriptor but this could be implemented [insert_idle](https://docs.rs/calloop/0.9.1/calloop/struct.LoopHandle.html#method.insert_idle).

Here are some other issues related to this bug:
https://github.com/rust-windowing/winit/issues/2063
https://github.com/NVIDIA/egl-wayland/issues/47

vberger:
Do I correctly gather that the new Nvidia driver spawns a thread a dispatches events from it, then? Those issues are multithread-specific. 🤔

HeavyRain266:
Looks like it does, since it's Nvidia on Linux, they do everything in _weird_ way. 🤔

Molytho:
Yes they do and I agree that it looks weird. They spawn a create a event queue and dispatch its events on another thread. It has something to do with their buffer freeing mechanism.
But nevertheless their implementation is allowed by the wayland protocol as far as I am aware of.

Dispatching queues on another thread is quite uncommon which is why this problem staid unnoticed for so long.

[...]

PR #213, opened 2021-12-26
wip: wayland-rs 0.30.0 rewrite

i509VCB:
This is the smithay client toolkit rewrite for the 0.30 wayland-rs api.

Given the large changes in wayland-rs this is effectively a rewrite.

Right now nothing is final and sctk is really just a way to battle test the new Wayland api.

I will edit this as the api becomes more concrete to describe the changes.


i509VCB:
One note, we should probably ensure we don't regress here assuming the calloop integration is kept around https://github.com/Smithay/client-toolkit/issues/212

src/registry.rs
```diff
@@ -0,0 +1,208 @@
...
+        match self.cached_globals.get(&name) {
+            Some(cached) => {
+                // Ensure the requested interface is the same.
+                if I::interface().name == cached.interface {
+                    // Create a new handle for the existing global.
+                    Ok(I::from_id(cx, cached.id.clone())?)
+                } else {
+                    Err(BindError::IncorrectInterface)
+                }
+            }
```
vberger, line 129: We need to handle the case where the underlying object is no longer alive (for example it was destroyed by something else). This is when `I::from_id` would return an error. In that case we should instead `log::warn!()` it and bind the global anew.

src/registry.rs
```diff
@@ -0,0 +1,208 @@
...
+                for handler in self.1.iter_mut() {
+                    handler.new_global(cx, qh, name, &interface[..], version, self.0);
+                }
+            }
+
+            wl_registry::Event::GlobalRemove { name } => {
+                for handler in self.1.iter_mut() {
+                    handler.remove_global(cx, name);
+                }
+            }
```
vberger, line 195: There is a possibility of global `name` reuse in the wayland protocol, as such when the `GlobalRemove` event is received, we should remove any associated global from the cache (but we don't need to destroy the wayland object itself), otherwise calls to `bind_cached` might erroneously fail.

src/shm/mod.rs
```diff
@@ -1,80 +1,155 @@
...
+use crate::registry::{RegistryHandle, RegistryHandler};
+
+use self::pool::{raw::RawPool, simple::SimplePool, CreatePoolError};
 
-pub use self::mempool::{AutoMemPool, DoubleMemPool, MemPool};
-pub use wl_shm::Format;
+pub trait ShmHandler {
+    /// Indicates the compositor supports an SHM pool with the specified format.
+    fn supported_format(&mut self, format: wl_shm::Format);
+}
```
vberger, line 16: I'm thinking the `ShmState` should instead store the supported formats in an internal `Vec`, and make it available to outside via a method, as was done previously. That's what the user is going to need anyway, no need to use this trait dance here, that is not an event the user is expected to need to answer to.

src/shm/pool/raw.rs
```diff
@@ -0,0 +1,235 @@
...
+        &mut self,
+        offset: i32,
+        width: i32,
+        height: i32,
+        stride: i32,
+        format: wl_shm::Format,
+        udata: U,
+        cx: &mut ConnectionHandle,
+        qh: &QueueHandle<D>,
+    ) -> Result<(&mut [u8], wl_buffer::WlBuffer), InvalidId>
```
vberger, line 83: I like this `RawPool` abstraction overall, but I'm not sure the `create_buffer` method should return a `&mut [u8]` slice to the underlying buffer. It's not necessary  for a low-level "raw" pool, and I'm afraid it will not interact well with higher-level constructs, and most notably buffer-reuse.

src/window/mod.rs
```diff
@@ -1,852 +1,600 @@
...
-struct WindowInner<F> {
-    frame: Rc<RefCell<F>>,
-    shell_surface: Arc<Box<dyn shell::ShellSurface>>,
-    user_impl: Box<dyn FnMut(Event, DispatchData)>,
-    min_size: (u32, u32),
-    max_size: Option<(u32, u32)>,
-    current_size: (u32, u32),
-    old_size: Option<(u32, u32)>,
-    decorated: bool,
+pub trait ShellHandler<D> {
```
vberger, line 51: I see this module is still very WIP, but overall it'd be nice if we manage to closely fit the API of libdecor, in terms of what shell-related methods are available, and what shell events are forwarded to the user. It'll make future integration with libdecor much easier.

For reference, the public API of libdecor is here: https://gitlab.gnome.org/jadahl/libdecor/-/blob/master/src/libdecor.h

src/output.rs
```diff
@@ -1,693 +1,674 @@
...
-    ///
-    /// If this is the case, it has become inert, you might want to
-    /// call its `release()` method if you don't plan to use it any
-    /// longer.
-    pub obsolete: bool,
+#[derive(Debug)]
+pub struct OutputState {
+    xdg: Option<ZxdgOutputManagerV1>,
+    outputs: Vec<OutputInner>,
 }
```
vberger, line 36: The `OutputState` should also provide a method that allows iterating over the outputs. And I think we would be better of with a freestanding function that allows to retrieve the `OutputInfo` from a `WlOutput`, and just expose the `WlOutput` to the user and let them get the info using that method if they need it.

src/output.rs
```diff
@@ -1,693 +1,674 @@
...
-    pub dimensions: (i32, i32),
-    /// Refresh rate for this mode, in mHz
-    pub refresh_rate: i32,
-    /// Whether this is the current mode for this output
-    pub is_current: bool,
-    /// Whether this is the preferred mode for this output
-    pub is_preferred: bool,
+    /// The info passed to this function was the state of the output before destruction.
+    fn output_destroyed(&mut self, info: OutputInfo);
 }
```
vberger, line 30: These methods need to pass the `WlOutput` object as well, of the user will not be able to do anything meaningful with them.

vberger:
Okay so, as discussed on Matrix, I pulled your commits into the [`rework-wayland-rs-0.30`](https://github.com/Smithay/client-toolkit/tree/rework-wayland-rs-0.30) branch of this repo. I suggest from this point we work by iterating on this branch with multiples PRs.

vberger:
Closing in favor of #215

PR #214, opened 2021-12-27
event_loop: correctly use prepare_read()

### LINKED ISSUE #212: Huge slowdowns with wayland on nvidia gpus

vberger:
Fixes #212

@Molytho can you confirm this one is correct for good?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+#### Bugfixes
+
+- SCTK now correctly interacts with the wayland socket being conccurently poolled from
+  other threads.
+
 ## 0.15.2 - 2021-10-27
 
 - Most types are now `Debug`
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ log = "0.4"
 wayland-client = "0.29"
 wayland-protocols = { version = "0.29" , features = ["client", "unstable_protocols"] }
 wayland-cursor = "0.29"
-calloop = { version = "0.9.1", optional = true }
+calloop = { version = "0.9.3", optional = true }
 
 [features]
 default = ["calloop", "dlopen"]
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -6,7 +6,7 @@ use calloop::{
     TokenFactory,
 };
 
-use wayland_client::EventQueue;
+use wayland_client::{EventQueue, ReadEventsGuard};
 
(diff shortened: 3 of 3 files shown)


Molytho:
It works perfectly fine 👍.

One last thing:
Maybe replace this https://github.com/Smithay/client-toolkit/blob/3c87c004812eefc559083b7c9a8000f4f884bf59/src/event_loop.rs#L99-L117
with that
https://github.com/Smithay/client-toolkit/blob/3c87c004812eefc559083b7c9a8000f4f884bf59/src/event_loop.rs#L167-L177
because `prepare_read` is no longer needed in the `Ok(_) => { ... }` path and because `prepare_read` can fail in the `Ok(0) => { ... } ` when something is read in between of `callback` and `prepare_read` (I know that's very unlikely to ever happen but when it happens once it's impossible to track down)

vberger:
Ah yes, good point!

Molytho:
It works. I think it's ready to be commited

PR #215, opened 2022-01-05
Rework for wayland-rs 0.30

vberger:
Roadmap from i5 for 0.30:

- [ ] Repeating source for keyboard implementation (#234)
- [ ] Cursor themes (#228)
- [ ] Client side decorations
- [ ] Data device
- [ ] Nth registry rework (#262)
- [ ] Shm pool abstractions
  - [x] MultiPool (#220)
  - [x] SlotPool (#241)
  - [ ] Buffer (#266)
- [ ] Write documentation


i509VCB:
Just going to note here to keep this idea in our minds.

Do we want to offer a sort of "super delegate type" for desktop like the `environment` preset in 0.29 era sctk.

One way this could be represented using a sort of `DesktopState` type which itself contains multiple delegate types (compositor, xdg shell, output, seat) and a single macro to delegate to that group of delegate types.

You would still need to implement some traits but it would make some very common groups of protocols easily usable.

i509VCB:
Do we want to consider more types of shm pools.

A double buffered pool is already essentially guaranteed to be implemented

There are some other possible useful pools implementations I can think of:
- SinglePool (name needs bikeshed) - A single shm buffer, meant for maximum control with race free access to the underlying content. This may be useful in very specific scenarios. This would not be implemented using a MultiPool as to reduce the overhead as much as possible.
- SwapchainPool (with name bikeshed) - Acts like a swapchain with N-buffering. This might be hard to justify since you would struggle to use the buffers faster than the server renders to the screen unless very aggressive damage tracking is used. Resizes could balloon really quickly due to the shear number of buffers. Might be useful for high refresh rate displays.

vberger:
Overall, I think those additional pools should not be considered as blocking the 0.30 update, and can be added after the merge.

i509VCB (approved):
This branch is being merged into master. All pending issues will have issues created and added to a milestone group.

PR #216, opened 2022-01-08
xdg: add delegate macro

i509VCB:


src/shell/xdg/window/mod.rs
```diff
@@ -66,7 +66,18 @@ pub enum CreateWindowError {
 }
 
 #[derive(Debug)]
-pub struct Window(pub(crate) Arc<WindowInner>);
+pub struct Window {
+    pub(crate) inner: Arc<WindowInner>,
+
+    /// Whether this is the primary handle to the window.
+    pub(crate) primary: bool,
```
vberger, line 73: COuld you expand this comment to explain why this primary mechanism is needed and what its role is?

PR #217, opened 2022-01-16
Multipool

snakedye:
Mempool that frees the buffer when release and favourizes buffer reuse.


src/shm/pool/multi.rs
```diff
@@ -0,0 +1,159 @@
...
+}
+
+impl<I: PartialEq + Clone> DelegateDispatchBase<wl_buffer::WlBuffer> for MultiPool<I> {
+    type UserData = ();
+}
+
+impl<D, I: PartialEq + Clone> DelegateDispatch<wl_buffer::WlBuffer, D> for MultiPool<I>
+where
+    D: Dispatch<wl_buffer::WlBuffer, UserData = Self::UserData>,
+    D: AsMut<MultiPool<I>>
```
i509VCB, line 143: This introduces an undue restriction on a number of pools you can have. For example, a pool per window is probably fine. I imagine the solution here is a sort of user data on the buffer that signals the pool it was freed.

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,159 @@
...
+    fn event(
+        data: &mut D,
+        buffer: &wl_buffer::WlBuffer,
+        event: wl_buffer::Event,
+        _: &Self::UserData,
+        conn: &mut ConnectionHandle,
+        _: &QueueHandle<D>
+    ) {
+        if let wl_buffer::Event::Release = event {
+            buffer.destroy(conn);
```
i509VCB, line 154: This destroys the object on the server, meaning reuse is impossible.

src/shm/pool/raw.rs
```diff
@@ -33,7 +33,7 @@ use super::CreatePoolError;
 #[derive(Debug)]
 pub struct RawPool {
     pool: wl_shm_pool::WlShmPool,
-    len: usize,
+    pub(crate) len: usize,
```
i509VCB, line 36: This could be a function for users of raw pool

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,159 @@
...
+impl<I: PartialEq + Clone> MultiPool<I> {
+    /// Resizes the memory pool, notifying the server the pool has changed in size.
+    ///
+    /// The wl_shm protocol only allows the pool to be made bigger. If the new size is smaller than the
+    /// current size of the pool, this function will do nothing.
+    pub fn resize(&mut self, size: usize, conn: &mut ConnectionHandle) -> io::Result<()> {
+        self.inner.resize(size, conn)
+    }
+    fn free(&self, buffer: &wl_buffer::WlBuffer) {
+        if let Some(buffer) = self.buffer_list.iter().find(|b| b.buffer.eq(buffer)) {
```
i509VCB, line 50: .eq -> `==`

i509VCB (changes requested):

src/shm/pool/multi.rs
snakedye: I'm not sure about the meaning of undue restrictions. Is it too much or too little restrictions? I added this so n buffering would be possible with multiple surfaces on the same mempool depending on the identifier. If it's not necessary I'll revert back.

i509VCB: AsMut contextually cannot select the right pool if you have more than one.

src/shm/pool/multi.rs
i509VCB: Well if the user is unable to reuse the buffer we gain no optimization here.

snakedye: The underlying buffer is the same it just has a different handle.

src/shm/pool/multi.rs
snakedye: I see. Maybe a trait like AsMutFrom<T, U> with U being the WlBuffer would work? This way the user could find the pool which holds the buffer.

src/shm/pool/multi.rs
snakedye: Iterating over the slice should be fine since the compositor can't create a wl_buffer with the same id if another is still in use.

src/shm/pool/multi.rs
snakedye: since buffer is a reference as b.buffer isn't it's cleaner this way. the other option would be to clone buffer.

vberger: I think `&b.buffer == buffer` is more direct for the eye (`==` is easily recognizable).

vberger:
Okay so, sorry for my late reply. I'll admit I've having a hard time understanding the logic of your `Multipool` (but maybe it's just because I'm tired). Could you give a general explanation of how it's meant to be used?

snakedye:
When a user wants to create a buffer they pass an identifier I, if the pool doesn't hold a buffer associated to that identifier, it will create it. The idea is to always return the buffer associated with the identifier to maximize buffer reuse.

Let's say I have multiple surfaces and want each of them to have a buffer. I'd create a Multipool<WlSurface> meaning the identifier of each Buffer is a WlSurface.

I can then use `create_buffer` like this
```rust
       pool.create_buffer(
            width,
            stride,
            height,
            surface, // One of our WlSurface
            FORMAT,
            udata,
            conn,
            qh,
      )
```
I do this for your two surfaces. Since there were no buffer with these surface in the pool, they are created. The pool looks somewhat like this now.

```rust
MultiPool {
    buffer_list: [
        Buffer {
            free: true,
            size: 231088,
            offset: 0,
            buffer: WlBuffer {
                id: ObjectId(wl_buffer@22),
                version: 1,
                data: Some(
                    ObjectData { .. },
                ),
            },
            identifier: WlSurface {
                id: ObjectId(wl_surface@12),
                version: 4,
                data: Some(
                    ObjectData { .. },
                ),
            },
        },
        Buffer {
            free: false,
            size: 231088,
            offset: 462176,
            buffer: WlBuffer {
                id: ObjectId(wl_buffer@13),
                version: 1,
                data: Some(
                    ObjectData { .. },
                ),
            },
            identifier: WlSurface {
                id: ObjectId(wl_surface@16),
                version: 4,
                data: Some(
                    ObjectData { .. },
                ),
            },
        },
    ],
    inner: ...
}
```
Now each time I use `create_buffer` on that pool with a WlSurface, the pool will try to return me the buffer at the same offset of the one with the matching WlSurface already in the pool. This way I'll always draw on the same buffer. 

Sometimes it happens that the buffer is moved. The `create_buffer` method returns the offset of the buffer so I can know if it moved. If it moved then I need to redraw everything. If it doesn't I can damage parts of it.

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,192 @@
...
+                if buffer.free.load(Ordering::Relaxed) {
+                    buffer.offset = offset;
+                } else {
+                    index = None;
+                }
+            } else if found_identifier {
+                break;
+            }
+            offset += buffer.size;
+        }
```
vberger, line 111: Could you add some comments explaining what this loop does? It's really not clear to me why this is changing the offset of others existing buffers when trying to create a buffer.

snakedye:
Hopefully the comments help, also in remove that was indeed an error.

vberger (commented):
Okay, this is taking a good shape I think. However I really think it'd be good to have some module-level documentation with an example of how to use this, with an explanation of the index keys and how that relates to double (or triple) buffering.

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,203 @@
...
+                    buffer.offset = offset;
+                } else {
+                    // If one of the overlapping buffers is busy, then no buffer can be returned because it could result in a data race.
+                    index = None;
+                }
+            } else if found_key {
+                break;
+            }
+            offset += buffer.size;
+        }
```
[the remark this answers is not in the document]
snakedye: > It seems to me that this is actually fine, because the method will just come back and fix its state on the next invocation, meaning that when it returns Some(...), the state is actually correct. Is that indeed your intent?

Yes. I'll comment that. Also I found [this](http://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/documentation.html) for documentation and it seems to cover everything I need but is there a way to preview the actual web page without compulsively running `cargo doc`?

> Then at a later point it redraws B, and then later again it tries to redraw its main surface, and A is still unusable because B is again in use. Following that pattern we can imagine unlucky cases where some buffers in the pool will never be usable again and end up as dead space.

I think if the client puts a cap of how much buffers they want, they can use frame callbacks to schedule another draw and by then I really doubt the buffer will still be in use. Worse case scenario, if the client doesn't have a hard cap, it should at least know when there's too much buffers for a surface and remove them manually.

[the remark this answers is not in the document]
vberger: > is there a way to preview the actual web page without compulsively running `cargo doc`?

I don't know about that sorry, I just run `cargo doc --no-deps --open` when I need to check how my documentation looks.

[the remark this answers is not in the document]
i509VCB: cargo-watch appears to exist for live reload, but I haven't used it before.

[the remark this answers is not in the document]
snakedye: I'll give it a shot, thanks.

vberger:
Okay, this is taking a good shape I'd say.

Just regarding this point:

> Sometimes it happens that the buffer is moved. The create_buffer method returns the offset of the buffer so I can know if it moved. If it moved then I need to redraw everything. If it doesn't I can damage parts of it.

Do I understand correctly that in practice you remember for each surface what was the last offset returned by the multipool when you requested a buffer, and use that to detect if you can do incremental draw or not? I think that's something that would be worth explaining in the documentation. As of now it is not really explained what this `offset` return value is.

Other than that small point, LGTM. :+1:

i509VCB:
Merged via #224 and superseded by #220

PR #220, opened 2022-02-14
Multipool (rebased)

vberger:
This is a rebased version of #217

Could you make sure I did not make any rebase mistake @snakedye , @i509VCB ?

### DIFF
--- a/src/shm/pool/mod.rs
+++ b/src/shm/pool/mod.rs
@@ -2,7 +2,7 @@ pub mod raw;
 pub mod multi;
 pub mod simple;
 
-use std::io;
+use std::{io, slice::Iter};
 
 use nix::errno::Errno;
 use wayland_client::backend::InvalidId;
@@ -34,5 +34,6 @@ impl From<Errno> for CreatePoolError {
 pub enum PoolHandle<'m, P> {
     Ref(&'m P),
     Slice(&'m [P]),
-    Vec(Vec<&'m P>)
+    Vec(Vec<&'m P>),
+    Iter(Iter<'m, P>)
 }
--- a/src/shm/pool/mod.rs
+++ b/src/shm/pool/mod.rs
@@ -38,5 +38,5 @@ pub trait AsPool<P> {
 pub enum PoolHandle<'m, P> {
     Ref(&'m P),
     Slice(&'m [P]),
-    RefSlice(&'m [&'m P])
+    Vec(Vec<&'m P>)
 }
--- a/src/shm/pool/mod.rs
+++ b/src/shm/pool/mod.rs
@@ -29,10 +29,6 @@ impl From<Errno> for CreatePoolError {
     }
 }
 
-pub trait AsPool<P> {
-    fn pool_handle(&self) -> PoolHandle<P>;
-}
-
 #[derive(Debug)]
 /// A handle to the underlying mempool
 pub enum PoolHandle<'m, P> {
--- a/src/shm/pool/mod.rs
+++ b/src/shm/pool/mod.rs
@@ -1,4 +1,5 @@
 pub mod raw;
+pub mod multi;
 pub mod simple;
 
 use std::io;
--- a/src/shm/pool/mod.rs
+++ b/src/shm/pool/mod.rs
@@ -1,5 +1,5 @@
-pub mod raw;
 pub mod multi;
+pub mod raw;
 pub mod simple;
 
 use std::io;
--- a/src/shm/pool/mod.rs
+++ b/src/shm/pool/mod.rs
@@ -2,7 +2,7 @@ pub mod raw;
 pub mod multi;
 pub mod simple;
 
-use std::{io, slice::Iter};
+use std::io;
 
 use nix::errno::Errno;
 use wayland_client::backend::InvalidId;
@@ -27,13 +27,4 @@ impl From<Errno> for CreatePoolError {
     fn from(errno: Errno) -> Self {
         Into::<io::Error>::into(errno).into()
     }
-}
-
-#[derive(Debug)]
-/// A handle to the underlying mempool
-pub enum PoolHandle<'m, P> {
-    Ref(&'m P),
-    Slice(&'m [P]),
-    Vec(Vec<&'m P>),
-    Iter(Iter<'m, P>)
 }
--- a/src/shm/pool/mod.rs
+++ b/src/shm/pool/mod.rs
@@ -28,3 +28,15 @@ impl From<Errno> for CreatePoolError {
         Into::<io::Error>::into(errno).into()
     }
 }
+
+pub trait AsPool<P> {
+    fn pool_handle(&self) -> PoolHandle<P>;
+}
+
+#[derive(Debug)]
+/// A handle to the underlying mempool
+pub enum PoolHandle<'m, P> {
+    Ref(&'m P),
+    Slice(&'m [P]),
+    RefSlice(&'m [&'m P])
+}
--- a/src/shm/pool/multi.rs
+++ b/src/shm/pool/multi.rs
@@ -1,9 +1,9 @@
 //! A pool implementation which automatically manage buffers.
 //!
-//!	This pool is built on the [`RawPool`].
+//! This pool is built on the [`RawPool`].
 //!
-//!	The [`MultiPool`] takes a key which is used to identify buffers and tries to return the buffer associated to the key
-//!	if possible. If no buffer in the pool is associated to the key, it will create a new one.
+//! The [`MultiPool`] takes a key which is used to identify buffers and tries to return the buffer associated to the key
+//! if possible. If no buffer in the pool is associated to the key, it will create a new one.
 //!
 //! # Example
 //!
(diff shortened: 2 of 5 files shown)


src/shm/pool/multi.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+impl<K: PartialEq + Clone> DelegateDispatchBase<wl_buffer::WlBuffer> for MultiPool<K> {
+    type UserData = ();
+}
+
+impl<D, K: PartialEq + Clone> DelegateDispatch<wl_buffer::WlBuffer, D> for MultiPool<K>
+where
+    D: Dispatch<wl_buffer::WlBuffer, UserData = Self::UserData>,
+    D: for<'m> AsPool<MultiPool<K>>
```
i509VCB, line 251: Not sure what has introduced the hrtb here?

src/shm/pool/mod.rs
```diff
@@ -27,3 +28,15 @@ impl From<Errno> for CreatePoolError {
         Into::<io::Error>::into(errno).into()
     }
 }
+
+pub trait AsPool<P> {
+    fn pool_handle(&self) -> PoolHandle<P>;
```
i509VCB, line 33: Let me think about this.

i509VCB (changes requested):

src/shm/pool/mod.rs
snakedye: I don't think this is needed anymore if I do this
```rust
impl<D, K: PartialEq + Clone> DelegateDispatch<wl_buffer::WlBuffer, D> for MultiPool<K>
where
    D: Dispatch<wl_buffer::WlBuffer, UserData = Self::UserData>,
    for<'p> &'p mut D: Into<PoolHandle<'p, MultiPool<K>>>
{
...
```

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,281 @@
...
+    fn event(
+        data: &mut D,
+        buffer: &wl_buffer::WlBuffer,
+        event: wl_buffer::Event,
+        _: &Self::UserData,
+        conn: &mut ConnectionHandle,
+        _: &QueueHandle<D>
+    ) {
+        if let wl_buffer::Event::Release = event {
+            buffer.destroy(conn);
```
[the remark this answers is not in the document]
snakedye: I didn't see it from the compositor pov. So the buffer should only be destroyed  when it's moved or resized? This adds a bit more complexity to an already complex pool. I'll see what I can do.

src/shm/pool/mod.rs
vberger: An other option is to use the `AsMut` trait, that might better reflect the intent

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,281 @@
...
+
+        Some((offset, buffer, slice))
+    }
+}
+
+impl<K: PartialEq + Clone> DelegateDispatchBase<wl_buffer::WlBuffer> for MultiPool<K> {
+    type UserData = ();
+}
+
+impl<D, K: PartialEq + Clone> DelegateDispatch<wl_buffer::WlBuffer, D> for MultiPool<K>
```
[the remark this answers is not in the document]
vberger: Yes, that's definitely an option, however there is the limitation that the lower-level API client-side cannot access the shared `D`, so the logic needs to be adjusted accordingly.

src/shm/pool/multi.rs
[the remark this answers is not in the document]
vberger: Not destroying the buffer is definitely more friendly to the compositor, but I'm pretty sure compositors can still do damage tracking even if the buffer is destroyed and recreated. However, I'm not sure about what compositors actually *do*.

src/shm/pool/mod.rs
snakedye: The issue with AsMut is that the return type must be a reference and I don't think the reference would live long enough or there would be a mismatch which means lifetime hell.

i509VCB: Let's consider an alternative for a bit before we decide which trait flavor is best.

The reason for this trait requirement is looking up the pool at runtime in order to modify internal state. Could we store a inner type of the pool `Arc/Weak<Mutex<InnerMultiPool>>` inside the user data and avoid the need for these lookup functions?

snakedye: I'd rather not use a reference counted pointer when it can be solved with a normal reference. It could also limit how the user wants to organize their pools.

i509VCB: As much as I agree with you about not wanting yet more reference counted pointers, wayland is a shared access mess sadly.

Considering this pool is arguably a mid-level pool (a semi double buffering pool needs to be built above this at some point), needing to expose a trait that returns a slice of MultiPool may not exactly work with the higher level abstractions. The shared pointer isn't ideal but it would only ever be internal to the wl_buffer and the pool.

I would think the destruction logic would be the following for buffers:

If the pool is first, destroy all wl_buffers that are not active (or at least queue it up since that needs &mut ConnectionHandle).
When active pools are sent the release event, if the pool has been dropped, destroy the wl_buffer immediately. The reference count hitting zero will finally destroy the inner pool.

snakedye: I am not sure I understand why it would be preferable to have an inner type. What would the InnerMultiPool look like?

I imagined a pool built on top of this one would own it so it wouldn't need to go use `PoolHandle`. 

> needing to expose a trait that returns a slice of MultiPool may not exactly work with the higher level abstractions. 

It returns a `PoolHandle` which is an enum. We could add the inner type to it. We'd still have to use Into to convert the Dispatch but it leaves a lot more option open.

i509VCB:
This has been merged via #224

vberger:
No, #224 has been merged into this PR 😅

i509VCB:
Well I guess I was on autopilot...

vberger:
Okay, I've added a few things to the PR. How does that look to you @i509VCB @snakedye ?

snakedye:
Looks good :)

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,295 @@
...
+    /// When it's not possible to use the buffer associated with the key, None is returned.
+    pub fn create_buffer(
+        &mut self,
+        width: i32,
+        stride: i32,
+        height: i32,
+        key: &K,
+        format: wl_shm::Format,
+        conn: &mut ConnectionHandle,
+    ) -> Option<(usize, wl_buffer::WlBuffer, &mut [u8])>
```
i509VCB, line 144: I feel it would be more ergonomic to split this function in two.

The first function would have a signature like this one to create the buffer, associating a key with the buffer slot.

The second function would act like an `acquire` function of sorts, returning a reference to the buffer and the slice representing it's content.

Of course this splitting would mean we need a way to resize a specific buffer in there based on it's key if it isn't in use.

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,295 @@
...
+    pub fn resize(&mut self, size: usize, conn: &mut ConnectionHandle) -> io::Result<()> {
+        self.inner.resize(size, conn)
+    }
+    /// Removes the buffer with the given key from the pool and rearranges the others
+    pub fn remove(&mut self, key: &K, conn: &mut ConnectionHandle) {
+        if let Some((i, buffer)) = self.buffer_list.iter().enumerate().find(|b| b.1.key.eq(key)) {
+            let mut offset = buffer.offset;
+            self.buffer_list.remove(i);
+            for buffer_handle in &mut self.buffer_list {
+                if buffer_handle.offset > offset && buffer_handle.free.load(Ordering::Relaxed) {
```
i509VCB, line 119: if the buffer is in use, then the buffer will leak.

We need a second field on the data of the buffer to tell it the pool has been destroyed and that it should release itself.

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,295 @@
...
+/// This pool manages buffers associated with keys.
+/// Only one buffer can be attributed to a given key.
+#[derive(Debug)]
+pub struct MultiPool<K: PartialEq + Clone> {
+    buffer_list: Vec<BufferHandle<K>>,
+    pub(crate) inner: RawPool,
+}
+
+#[derive(Debug)]
+struct BufferHandle<K: PartialEq + Clone> {
```
i509VCB, line 90: nit: probably better to name this something like `BufferAllocation` or `BufferSlot`.

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,295 @@
...
+    free: Arc<AtomicBool>,
+    size: usize,
+    used: usize,
+    offset: usize,
+    buffer: Option<wl_buffer::WlBuffer>,
+    key: K,
+}
+
+impl<E: PartialEq + Clone> From<RawPool> for MultiPool<E> {
+    fn from(inner: RawPool) -> Self {
```
i509VCB, line 100: We have no way to guarantee we wouldn't be stomping over existing wl_shm allocations that could come from the raw pool if someone else did something with the pool before we were given it.

We should replace this with a proper constructor that doesn't expose the raw pool to the public api.

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,295 @@
...
+        Self { buffer_list: Vec::new(), inner }
+    }
+}
+
+impl<K: PartialEq + Clone> MultiPool<K> {
+    /// Resizes the memory pool, notifying the server the pool has changed in size.
+    ///
+    /// The wl_shm protocol only allows the pool to be made bigger. If the new size is smaller than the
+    /// current size of the pool, this function will do nothing.
+    pub fn resize(&mut self, size: usize, conn: &mut ConnectionHandle) -> io::Result<()> {
```
i509VCB, line 110: We should note this performs a manual resize, such as in the case where you don't want to tell the server you've resized 20 times. Also note creating a new buffer will resize the pool for you

i509VCB (changes requested):

snakedye:
https://github.com/Smithay/client-toolkit/blob/de6546c193d9abc71ffd012339da2b4bcf6954c6/src/shm/pool/multi.rs#L117 the buffer isn't destroyed when it's removed.

src/shm/pool/multi.rs
vberger: I'm not sure I get your point. Could you explain more how it would improve the ergonomics, and how it'd be used?

src/shm/pool/multi.rs
vberger: I'm not sure about that, isn't it the same delayed reorganizing as was discussed there?

https://github.com/Smithay/client-toolkit/pull/217#discussion_r795088748

src/shm/pool/multi.rs
i509VCB: One way to look at a multi pool is a sort of `HashMap<K, WlBuffer> that guards lookup of an entry if the server has not released the buffer.

Essentially my idea above was to make the semanatics of the equivalentish insert, get, remove, etc operations of the multi pool more explicit to a user of the api.

Currently the smenatics of get, insert and resize all correspond to the `create_buffer` function.

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,397 @@
...
+    where
+        Q: Eq,
+        K: std::borrow::Borrow<Q>,
+    {
+        if let Some((i, buf)) =
+            self.buffer_list.iter().enumerate().find(|(_, slot)| slot.key.borrow().eq(key))
+        {
+            let mut offset = buf.offset;
+            if let Some(buf) = self.buffer_list.remove(i).buffer {
+                buf.destroy(conn);
```
i509VCB, line 133: This may destroy a buffer in use

snakedye: I think this if let block can be replaced by
```rust
self.buffer_list.remove(i);
```
The buffer will not be destroyed so it might not be ideal.

I think you would only remove a buffer if you were done with it ie the surface is destroyed. In that case, I don't think it's a big deal to potentially destroy the buffer since the compositor should not display anything on it in the following frame.

vberger:
closing in favor of #257

PR #225, opened 2022-02-19
xdg-shell (0.30): split windowing out of XdgShellState

i509VCB:
This splits the windows out of `XdgShellState` and into a new `XdgWindowState`. This makes the xdg shell abstractions much more modular and removes the need to have `delgate_xdg_shell` require `delegate_xdg_window`.

### DIFF
--- a/src/shell/xdg/window/mod.rs
+++ b/src/shell/xdg/window/mod.rs
@@ -1,18 +1,18 @@
-use std::sync::{
-    atomic::{AtomicBool, Ordering},
-    Arc,
-};
+use std::sync::{atomic::AtomicBool, Arc};
 
 use wayland_backend::client::InvalidId;
 use wayland_client::{
     protocol::{wl_output, wl_surface},
-    ConnectionHandle, Dispatch, Proxy, QueueHandle,
+    ConnectionHandle, Dispatch, QueueHandle,
 };
 use wayland_protocols::{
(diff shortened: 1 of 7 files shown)


src/shell/xdg/window/mod.rs
```diff
@@ -9,28 +9,124 @@ use wayland_client::{
...
+            + Dispatch<xdg_toplevel::XdgToplevel, UserData = WindowData>
+            + 'static,
+    {
+        // We don't know if an un-managed surface has a role object.
+        let surface_data = wl_surface.data::<SurfaceData>().ok_or(CreateWindowError::HasRole)?;
+
+        // XDG Shell protocol forbids creating an window from a surface that already has a role object.
+        if surface_data.has_role.load(Ordering::SeqCst) {
+            return Err(CreateWindowError::HasRole);
+        }
```
vberger, line 99: I'm not quite sure about those checks, they seem unnecessary as the compositor will check this anyway and trigger a protocol error if necessary, and it seems to me that this kind of condition is *always* a logic bug, so having the function return and error is kind of pointless is the user is always going to `.unwrap()` it anyway. Crashing on the protocol error achieves the same result.

This applies to the whole "SCTK tracking the surface roles itself" actually, not just that particular instance.

i509VCB: Considering I've changed the logic to consume the surface, it might not be too nessecary to check this.

vberger (approved):

PR #229, opened 2022-03-31
Add touch support

### LINKED ISSUE #169: Touch input
If I understood correctly, now it's only possible to check whether seat has touch support or not, without any interface for handling input. It would be nice to support it.

Weston client as a usage example: https://github.com/wayland-project/weston/blob/master/clients/simple-touch.c

Protocol [retrieving an ojbect](https://github.com/wayland-project/wayland/blob/master/protocol/wayland.xml#L1819-L1831) and its [interface](https://github.com/wayland-project/wayland/blob/master/protocol/wayland.xml#L2304-L2446) (anything else missed?).

chrisduerr:
This adds a new handler for touch events. Since most touch events are
buffered until a frame event is sent, this buffering is performed
internally similar to the pointer implementation.

The implementation supports wl_touch up to version 6, including the
touch point shape feature.

Fixes #169.


chrisduerr:
As discussed in #169, there isn't really much that needs to be done by SCTK itself for supporting touch input. However it does manage to abstract away frame event handling which one would likely otherwise find in every single client handling touch input.

There's also the argument to be made that this helps with documentation and discoverability. The current 0.30 examples don't really go into great detail on how to handle things not implemented directly in SCTK (though it is fairly trivial) and one might struggle to figure out that this is what they need to do when they want to add touch support, given that pointer and keyboard is directly in SCTK.

src/seat/touch.rs
```diff
@@ -0,0 +1,184 @@
...
+        conn: &mut ConnectionHandle,
+        qh: &QueueHandle<Self>,
+        touch: &WlTouch,
+        serial: u32,
+        time: u32,
+        surface: WlSurface,
+        id: i32,
+        x: f64,
+        y: f64,
+    );
```
i509VCB, line 53: I kind of agree with clippy here. A non-exhaustive struct should satisfy this while allowing future extension (if it happens) to occur.

i509VCB (approved):

src/seat/touch.rs
chrisduerr: I've wanted to follow the behavior of the other handlers here and largely mimic the events and break everything out.

Starting to bundle things into structs for one API but not the others seems very inconsistent to me? Especially with things like `serial` having it as a parameter in one and a subparameter of some random parameter in another would be hugely irritating.

vberger: I think I agree to keep consistency with the rest of the API, introducing a struct for this event only would be too confusing.

I guess an alternative option would be to have a `TouchPointEvent` enum that fuses `down`, `up`, `motion`, `shape` and `orientation` into a single `fn (&mut self, &mut ConnectionHandle, &QueueHandle<Self>, WlTouch, TourchPointEvent)` method.

src/seat/touch.rs
```diff
@@ -0,0 +1,184 @@
...
+                guard.events.push(event);
+            }
+            // Process all buffered events.
+            TouchEvent::Frame => {
+                let mut guard = udata.inner.lock().unwrap();
+                for event in guard.events.drain(..) {
+                    process_framed_event(data, touch, conn, qh, event);
+                }
+            }
+            TouchEvent::Cancel => data.cancel(conn, qh, touch),
```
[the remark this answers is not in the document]
chrisduerr: You mean on cancel? I'm not sure to be honest.

Generally I don't think the cancel event is tied to frame events since it's created by the compositor. What I would assume is that a cancel is generally only transmitted after all relevant events have been framed, since only on frame should they actually be processed to be recognized as a gesture?

Like the frame isn't meant to be used to buffer touch input that isn't detected as gesture yet, but more as a debounce mechanism? Though I suppose it probably shouldn't *hurt* to just clear it when we assume all good-behaving compositors will never call cancel with a pending frame event?

src/seat/touch.rs
```diff
@@ -0,0 +1,184 @@
...
+    /// Touch point motion.
+    #[allow(clippy::too_many_arguments)]
+    fn motion(
+        &mut self,
+        conn: &mut ConnectionHandle,
+        qh: &QueueHandle<Self>,
+        touch: &WlTouch,
+        time: u32,
+        id: i32,
+        x: f64,
```
i509VCB, line 75: Do we want to note what coordinate system the x and y are in?

src/seat/touch.rs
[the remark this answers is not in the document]
vberger: > Though I suppose it probably shouldn't hurt to just clear it when we assume all good-behaving compositors will never call cancel with a pending frame event?

Yeah, that's mostly what I have in mind

src/seat/touch.rs
chrisduerr: I don't think a single handler is the way to go tbh. Because that way you're splitting things up, then you match them in SCTK, just to combine them again and have the client match it again. That just seems unfortunate for downstream and at that point causes more harm than it does good.

There are some minor things one can do, combining x and y to a tuple is done in other APIs for example and iirc that might remove one of the warnings. Other than that it's rough though with touch down itself, there's lots to unpack there wrt information contained in it.

Besides the clippy ignore, which I don't really like myself, I think the best path would probably just be to keep frequently used information flat and obstruct infrequently used information. Though that highly depends on the usecase. One person might think `time` is mostly useless while another treats every `id` the same anyway.

chrisduerr: I've removed one of the clippy annotations, combining coordinates similar to what is done in `pointer.rs`. But I don't think there's a great way to do that for the `down` function without transforming all other functions too and creating something inconsistent to the other APIs.

So I've just left the annotation on there for now. It's certainly not ideal but I think it's okay.

src/seat/touch.rs
```diff
@@ -38,7 +38,9 @@ pub trait TouchHandler: SeatHandler + Sized {
     ///
     /// The associated touch ID ceases to be valid after the touch up event with the associated ID
     /// and may be reused for other touch points after that.
-    #[allow(clippy::too_many_arguments)]
+    ///
+    /// Coordinates are surface-local.
+    #[warn(clippy::too_many_arguments)]
```
vberger, line 43: I don't think we need to keep that

vberger: I mean, this lint is already warn by default, why explicitly enable it again?

chrisduerr: Ah, I meant to `allow` it, not `warn` on it.

vberger (approved):

PR #231, opened 2022-04-03
0.30 xdg: idea to make configures better

i509VCB:
Vast improvements to `Window` abstraction in xdg shell handling for SCTK.

Changes include:
- Moving configure to `WindowHandler`, each derivation of `XdgSurface` is expected to provide it's own handler method for configures.
- Window creation uses a builder
- Creating an XdgSurface, returned as a `XdgShellSurface` is now possible to allow external implementations of protocols which extend `XdgSurface`.
- Better server side decoration handling
- General cleanup of window data

### DIFF
--- a/src/shell/xdg/inner.rs
+++ b/src/shell/xdg/inner.rs
@@ -5,11 +5,9 @@ use wayland_protocols::xdg_shell::client::{xdg_surface, xdg_wm_base};
 
 use crate::registry::{ProvidesRegistryState, RegistryHandler};
 
-use super::{XdgShellHandler, XdgShellState};
+use super::{XdgShellHandler, XdgShellState, XdgSurfaceData};
 
-pub(crate) const MAX_XDG_WM_BASE: u32 = 3;
-
-impl<D> RegistryHandler<D> for XdgShellState
+impl<D> RegistryHandler<D> for XdgShellState<D>
 where
     D: Dispatch<xdg_wm_base::XdgWmBase, UserData = ()>
         + XdgShellHandler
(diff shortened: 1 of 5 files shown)


src/shell/xdg/inner.rs
```diff
@@ -41,26 +36,18 @@ where
...
-            .xdg_shell_state()
-            .xdg_wm_base
-            .as_ref()
-            .filter(|(global_name, _)| global_name == &name)
-            .is_some()
-        {
-            todo!("XDG shell global destruction")
-        }
+    fn remove_global(_: &mut D, _: &mut ConnectionHandle, _: &QueueHandle<D>, _: u32) {
+        // TODO: How to handle "defunct surfaces" error
```
vberger, line 40: That's not really a problem tbh, removal of the `xdg_wm_base` global is definitely in the "will never happen" category.

src/shell/xdg/window/mod.rs
```diff
@@ -164,30 +135,227 @@ pub struct WindowConfigure {
...
+    ///
+    /// This may be used as a compositor hint to send an initial configure with the specified minimum size.
+    pub fn min_size(self, min_size: (u32, u32)) -> Self {
+        Self { min_size: Some(min_size), ..self }
+    }
+
+    /// Suggest the maximum size of the window being built.
+    ///
+    /// This may be used as a compositor hint to send an initial configure with the specified maximum size.
+    pub fn set_max_size(self, max_size: (u32, u32)) -> Self {
```
vberger, line 201:
```suggestion
    pub fn max_size(self, max_size: (u32, u32)) -> Self {
```

src/shell/xdg/mod.rs
```diff
@@ -1,58 +1,142 @@
...
+    MissingRequiredGlobals,
+
+    /// Protocol error.
+    #[error(transparent)]
+    Protocol(#[from] InvalidId),
+}
+
 #[derive(Debug)]
-pub struct XdgShellState {
+pub struct XdgShellState<D> {
```
vberger, line 26: Why is this type parameter needed? Wouldn't making `create_xdg_surface` generic instead work?

i509VCB:
```
error[E0207]: the type parameter `D` is not constrained by the impl trait, self type, or predicates
  --> src/shell/xdg/inner.rs:72:6
   |
72 | impl<D> DelegateDispatchBase<xdg_surface::XdgSurface> for XdgShellState
   |      ^ unconstrained type parameter
```

for
```rust
impl<D> DelegateDispatchBase<xdg_surface::XdgSurface> for XdgShellState {
    type UserData = XdgSurfaceData<D>;
}
```

vberger: Oh, right :+1:

vberger (approved):

PR #235, opened 2022-04-08
0.30: layer shell

i509VCB:
Shell abstractions for the layer shell. This is modeled off the xdg shell abstractions currently in 0.30.

We intentionally wrap types used by the wlr protocol since there is a future ext extension being proposed. We do not implement the ext extension (but it was considered during the design).

### DIFF
--- a/src/shell/layer/mod.rs
+++ b/src/shell/layer/mod.rs
@@ -1,0 +1,496 @@
+mod dispatch;
+
+use std::{
+    convert::TryFrom,
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+};
+
+use bitflags::bitflags;
+use wayland_backend::client::InvalidId;
+use wayland_client::{
+    protocol::{wl_output, wl_surface},
+    ConnectionHandle, Dispatch, QueueHandle,
+};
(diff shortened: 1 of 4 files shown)


src/shell/layer/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+    pub(crate) death_signal: Arc<AtomicBool>,
+}
+
+impl PartialEq for Layer {
+    fn eq(&self, other: &Self) -> bool {
+        self.kind == other.kind
+    }
+}
+
+impl Layer {
```
vberger, line 227: This represents a layer surface, not a layer. Let's name it `LayerSurface` instead (and similarly for the related types like the builder).

src/shell/layer/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+        Self::None
+    }
+}
+
+/// The z-depth of a layer.
+///
+/// These values indicate which order a [`Layer`] is rendered in.
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum ZDepth {
```
vberger, line 345: This, on the other hand, represents the layer, so this enum should be named `Layer`.

vberger (approved):

PR #239, opened 2022-04-20
DataOffer low level receive method & DataDevice with_dnd method

wash2:
This PR adds:
- a method to DataOffer which receives data to a raw file descriptor, then closes it.
- a method to DataDevice which gives access to the current DnD offer outside of a DnDEvent, which I've found useful.

I'm working on wrapping xdg windows with a layer shell window. One of the use cases I'd like to support is forwarding the clipboard contents into the embedded compositor inside the layer shell window. What I'm currently trying to do is:

1. set the compositor provided selection when the layer shell window gains keyboard focus for a seat. https://smithay.github.io/smithay/smithay/wayland/data_device/fn.set_data_device_selection.html

2. Wait for the SendSelection request in the callback I provided to init_data_device. https://docs.rs/smithay/latest/smithay/wayland/data_device/fn.init_data_device.html

3. Forward the provided fd from the SendSelection request (2) to the receive method of the data offer for the seat from (1)

Step 3 is what this PR is intended to make possible.

### DIFF
--- a/src/data_device/offer.rs
+++ b/src/data_device/offer.rs
@@ -106,6 +106,29 @@ impl DataOffer {
         Ok(unsafe { FromRawFd::from_raw_fd(readfd) })
     }
 
+    /// Receive data to the write end of a raw file descriptor. If you have the read end, you can read from it.
+    ///
+    /// You can do this several times, as a reaction to motion of
+    /// the dnd cursor, or to inspect the data in order to choose your
+    /// response.
+    ///
+    /// Note that you should *not* read the contents right away in a
+    /// blocking way, as you may deadlock your application doing so.
+    /// At least make sure you flush your events to the server before
+    /// doing so.
+    ///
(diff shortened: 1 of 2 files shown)


src/data_device/offer.rs
```diff
@@ -106,6 +106,29 @@ impl DataOffer {
...
+        use nix::unistd::close;
+
+        self.offer.receive(mime_type, writefd);
+
+        if let Err(err) = close(writefd) {
+            log::warn!("Failed to close write pipe: {}", err);
+        }
+
+        Ok(())
+    }
```
PolyMeilex, line 130: Is `std::io::Result<()>` the right return type for this? The method always returns `Ok(())`

wash2: Oh good point, my bad. I'll fix that.

PolyMeilex (approved):

vberger (approved):

vberger:
Thanks!

The failing CI is unrelated to this PR.

PR #241, opened 2022-05-01
Port AutoMemPool to 0.30-rework

danieldg:
This provides an API based on dividing the pool into slots.  It is simpler than MultiPool because it does not move slots, which might result in more memory use if there are many resize operations.

### DIFF
--- a/src/shm/pool/raw.rs
+++ b/src/shm/pool/raw.rs
@@ -150,21 +150,17 @@ impl io::Seek for RawPool {
 }
 
 impl RawPool {
-    pub(crate) fn new<D, U>(
-        len: usize,
-        shm: &wl_shm::WlShm,
-        qh: &QueueHandle<D>,
-        udata: U,
-    ) -> Result<RawPool, CreatePoolError>
-    where
-        D: Dispatch<wl_shm_pool::WlShmPool, UserData = U> + 'static,
-        U: Send + Sync + 'static,
-    {
+    pub(crate) fn new(len: usize, shm: &wl_shm::WlShm) -> Result<RawPool, CreatePoolError> {
         let shm_fd = RawPool::create_shm_fd()?;
         let mem_file = unsafe { File::from_raw_fd(shm_fd) };
         mem_file.set_len(len as u64)?;
 
-        let pool = shm.create_pool(shm_fd, len as i32, qh, udata).map_err(GlobalError::from)?;
+        let pool = shm
+            .send_constructor(
+                wl_shm::Request::CreatePool { fd: shm_fd, size: len as i32 },
+                Arc::new(ShmPoolData),
+            )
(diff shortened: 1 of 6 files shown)


danieldg:
This currently relies on https://github.com/Smithay/wayland-rs/pull/483

src/shm/pool/raw.rs
```diff
@@ -94,6 +101,33 @@ impl RawPool {
         Ok(buffer)
     }
 
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn create_buffer_raw(
```
i509VCB, line 105: is this something other pool implementations (out of crate implementations especially) could use and justify making this function public?

danieldg: I think so, although depending on how the client API works out, it may not be needed as long as the `pool()` function exists to allow direct access to the interface.

src/shm/pool/slot.rs
```diff
@@ -0,0 +1,501 @@
...
+
+/// A [`wl_buffer::WlBuffer`] allocated from a [SlotPool].
+///
+/// When this object is dropped, the buffer will be destroyed immediately if it is not active, or
+/// upon the server's release if it is.
+#[derive(Debug)]
+pub struct Buffer {
+    buffer: wl_buffer::WlBuffer,
+    height: i32,
+    stride: i32,
```
i509VCB, line 84: width could be useful if you need to tell a software renderer the extent of the framebuffer.

i509VCB (approved):
Everything else appears to look good.

src/shm/pool/slot.rs
danieldg: It's easy to track this yourself if it's not a trivial calculation from the stride (most commonly - ARGB8888 with align 1 or 4 - it will be stride/4). The values that are stored are just the ones needed to produce the correct buffer size.

snakedye: If format of the buffer is exposed you could determine the size but otherwise the user has to use a fixed format for all the buffers to know the amount of bytes per pixel. Most of the time I'll be 4 but sometimes it might not be. I don't know how much you value that but it'd be better if the pool was capable of supporting all formats.

danieldg: Tracking formats by storing the used format alongside the buffer seems reasonable to me.  If you support anything beyond {A,X}RGB8888 you already need to track support prior to creating the buffer, so you either store this alongside the pool or you can store it alongside the buffer.

Alternatively the `Buffer` struct could retain the full fields as configured, if this type of tracking is going to be normal for the SCTK front-ends for various protocol objects: having surfaces or windows track their size, scale, and transform, and other state.

vberger (approved):

PR #242, opened 2022-05-01
Expose globals via State objects

danieldg:
Most RegistryHandlers expose their contained global; add it to Shm and Compositor.


i509VCB (approved):
I swear I recall exposing the protocol objects...

PR #246, opened 2022-05-08
RFC: SlotPool shared memory API

danieldg:
This provides a simpler, more robust API than MultiPool.  Unlike MultiPool, it does not move slots, which might result in more memory use if there are many resize operations.

Based on #245 because it relies on calling destructors in `Drop::drop`.


src/shm/pool/slot.rs
```diff
@@ -0,0 +1,412 @@
...
+pub enum ActivateSlotError {
+    /// Buffer was already active
+    #[error("Buffer was already active")]
+    AlreadyActive,
+}
+
+#[derive(Debug)]
+pub struct SlotPool {
+    pub(crate) inner: RawPool,
+    free_list: Arc<Mutex<Vec<(usize, usize)>>>,
```
i509VCB, line 46: would be nice to know what the two usize values are

src/shm/pool/slot.rs
```diff
@@ -0,0 +1,412 @@
...
+const BUFFER_STATE_DESTROY_ON_RELEASE: u8 = 2;
+
+/// Buffer has been destroyed
+const BUFFER_STATE_DEAD: u8 = 3;
+
+/// Value that is ORed on buffer release to transition to the next state
+const BUFFER_STATE_RELEASE_SET: u8 = 1;
+
+/// Value that is ORed on buffer destroy to transition to the next state
+const BUFFER_STATE_DESTROY_SET: u8 = 2;
```
i509VCB, line 78: Might be nice to note these exist because we don't have any AtomicEnum

src/shm/pool/slot.rs
```diff
@@ -0,0 +1,412 @@
...
+        })
+    }
+
+    /// Get the bytes corresponding to a given slot.
+    ///
+    /// Note: prefer using canvas, which will prevent drawing to a buffer that has not been
+    /// released by the server.
+    ///
+    /// Returns an empty buffer if the slot does not belong to this pool.
+    pub fn raw_data_mut(&mut self, slot: &Slot) -> &mut [u8] {
```
i509VCB, line 242: I assume this is for higher level pools based on the slot pool?

src/shm/pool/slot.rs
```diff
@@ -0,0 +1,412 @@
...
+                offset,
+                len,
+                active_buffers: AtomicUsize::new(0),
+            }),
+        })
+    }
+
+    /// Get the bytes corresponding to a given slot.
+    ///
+    /// Note: prefer using canvas, which will prevent drawing to a buffer that has not been
```
i509VCB, line 238: let's use the doc comments to link to `SlotPool::canvas`

src/shm/pool/slot.rs
```diff
@@ -0,0 +1,412 @@
...
+        let canvas = self.raw_data_mut(&slot);
+        Ok((slot, buffer, canvas))
+    }
+
+    /// Get the bytes corresponding to a given slot if drawing to the slot is permitted.
+    ///
+    /// Returns `None` if there are active buffers in the slot or if the slot does not correspond
+    /// to this pool.
+    pub fn canvas(&mut self, slot: &Slot) -> Option<&mut [u8]> {
+        if slot.has_active_buffers() {
```
i509VCB, line 130: I could see it being useful to access the canvas of one specific buffers in a slot assuming it is not in use.

src/shm/pool/slot.rs
```diff
@@ -0,0 +1,412 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum ActivateSlotError {
+    /// Buffer was already active
+    #[error("Buffer was already active")]
+    AlreadyActive,
+}
+
+#[derive(Debug)]
+pub struct SlotPool {
+    pub(crate) inner: RawPool,
```
i509VCB, line 45: The shm pool is leaked as RawPool does not cleanup on drop.

In particular do note that a buffer **can** outlive it's pool protocol object

src/shm/pool/slot.rs
```diff
@@ -0,0 +1,412 @@
...
+const BUFFER_STATE_DEAD: u8 = 3;
+
+/// Value that is ORed on buffer release to transition to the next state
+const BUFFER_STATE_RELEASE_SET: u8 = 1;
+
+/// Value that is ORed on buffer destroy to transition to the next state
+const BUFFER_STATE_DESTROY_SET: u8 = 2;
+
+#[derive(Debug)]
+pub struct Buffer {
```
i509VCB, line 81: It would be quite nice to be able to use a buffer to get the area to draw into vs getting the entire slot's content and calculating where to draw myself

src/shm/pool/slot.rs
```diff
@@ -0,0 +1,412 @@
...
+                data.inner.active_buffers.fetch_add(1, Ordering::Relaxed);
+                Ok(())
+            }
+            Err(BUFFER_STATE_ACTIVE) => Err(ActivateSlotError::AlreadyActive),
+            _ => unreachable!("Invalid state in BufferData"),
+        }
+    }
+
+    /// Manually mark a buffer as inactive.
+    pub fn deactivate(&self) -> Result<(), ActivateSlotError> {
```
i509VCB, line 344: Elaborate on why someone shouldn't use this function in the docs

src/shm/pool/slot.rs
danieldg: Yes, or for things like using read-only access to a buffer that is in use by the server.

src/shm/pool/slot.rs
danieldg: A slot is either used or unused as a whole; it generally only has one buffer.  I do plan to add a `fn canvas(&mut self, buffer: &Buffer)` however.

src/shm/pool/slot.rs
danieldg: I think that cleanup should be done in `RawPool`, unless there's a reason not to?

danieldg: correction: RawPool already does cleanup on drop.  Buffers continue to exist (and are usable, although you cannot access their canvas anymore) until they are dropped and released.

src/shm/pool/slot.rs
```diff
@@ -0,0 +1,514 @@
...
+            return Err(CreateBufferError::PoolMismatch);
+        }
+
+        let data = Arc::new(BufferData {
+            inner: slot.inner.clone(),
+            state: AtomicU8::new(BufferData::INACTIVE),
+        });
+        let buffer =
+            self.inner.create_buffer_raw(offset, width, height, stride, format, data)?;
+        buffer.backend().upgrade().expect("alive").get_data(buffer.id()).expect("has").downcast_ref::<BufferData>().expect("is");
```
i509VCB, line 311: Better `.expect()` descriptions would be nice.

danieldg:
Replaced by updating #241

src/shm/pool/slot.rs
i509VCB, line 311: This statement appears to be some sort of assertion that is nearly always true after the `has` expect?

danieldg: This was a test line that was left in by accident.

PR #248, opened 2022-05-13
Update Calloop to 0.10.0

wash2:
This PR updates Calloop to 0.10.0

### DIFF
--- a/src/event_loop.rs
+++ b/src/event_loop.rs
@@ -1,12 +1,8 @@
-use std::{
-    io,
-    os::unix::io::RawFd
-};
+use std::{io, os::unix::io::RawFd};
 
 use calloop::{
-    generic::{Generic},
-    EventSource, InsertError, Interest, LoopHandle, Mode, PostAction, RegistrationToken,
-    TokenFactory,
+    generic::Generic, EventSource, InsertError, Interest, LoopHandle, Mode, PostAction,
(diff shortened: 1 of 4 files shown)


src/event_loop.rs
```diff
@@ -125,23 +121,23 @@ impl EventSource for WaylandSource {
         &mut self,
         poll: &mut calloop::Poll,
         token_factory: &mut TokenFactory,
-    ) -> std::io::Result<()> {
+    ) -> Result<(), calloop::Error> {
```
i509VCB, line 124: This should be io::Result?

src/event_loop.rs
```diff
@@ -171,7 +167,7 @@ impl EventSource for WaylandSource {
         Ok(())
     }
 
-    fn post_run<F>(&mut self, _: F) -> std::io::Result<()>
+    fn post_run<F>(&mut self, _: F) -> Result<(), calloop::Error>
```
i509VCB, line 170: calloop::Result<()> can be used here

src/event_loop.rs
wash2: when I try to use `io::Result`, I see an error. `calloop::Result` is fine though

vberger:
Given this is a breaking change of the API, I'd be inclined to integrating it as part of the 0.30 rework, rather than make a breaking calloop release for just that.

PR #249, opened 2022-05-19
layer_shell example: use Vec::retain_mut()

MaxVerevkin:
`Vec::retain_mut()` is stabilized in Rust 1.61

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust: ['1.42.0', 'stable', 'beta']
+        rust: ['1.61.0', 'stable', 'beta']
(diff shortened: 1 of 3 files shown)


i509VCB:
When is 1.61 released?

i509VCB:
Also I believe there is a similar function in xdg shell (window)

MaxVerevkin:
> Also I believe there is a similar function in xdg shell (window)

I couldn't find it with `rg 'while.*len' .`

vberger (approved):

i509VCB:
the ci step for 1.42 will need to be bumped to 1.61

also note the change in the readme and get the lint to pass.

MaxVerevkin:
> get the lint to pass

There are quite a lot of lints not related to my PR.

PR #251, opened 2022-05-22
Multipool: Use borrow type to find the key 

snakedye:
Adapt a similar behaviour to rust's hashmap where a borrowing type used to compare the keys.


vberger:
The FreeBSD error was spurious, I restarted the job. There is a nit rustfmt is unhappy about though.

Other than that, LGTM :+1:

PR #252, opened 2022-05-22
Custom user data

danieldg:
Allow applications to store data attached to wayland objects in addition to the data that SCTK stores there.

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -39,6 +39,16 @@ pub trait CompositorHandler: Sized {
     );
 }
 
+pub trait SurfaceDataExt: Send + Sync {
+    fn surface_data(&self) -> &SurfaceData;
+}
+
+impl SurfaceDataExt for SurfaceData {
+    fn surface_data(&self) -> &SurfaceData {
+        self
+    }
+}
+
 #[derive(Debug)]
 pub struct CompositorState {
     wl_compositor: GlobalProxy<wl_compositor::WlCompositor>,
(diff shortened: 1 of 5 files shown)


examples/image_viewer.rs
```diff
@@ -0,0 +1,356 @@
...
+                return;
+            }
+        };
+
+        // We'll need the image in RGBA for drawing it
+        let image = image.to_rgba8();
+
+        pool_size += image.width() * image.height() * 4;
+
+        let viewer = Arc::new_cyclic(|weak| {
```
i509VCB, line 71: What's the rationale behind storing all the image state in an arc and sharing it with every window vs using the &mut self reference you have in all the callbacks?

danieldg: Avoiding the requirement that the `&mut self` maintain and walk a list of all surfaces when handling events.

Looking at the individual patches instead of the big diff makes this clearer.

danieldg: I could drop the Demonstrate patch; the example might be clearer without it.

src/compositor.rs
```diff
@@ -103,55 +152,104 @@ macro_rules! delegate_compositor {
             ] => $crate::compositor::CompositorState
         );
     };
+    ($ty: ty, surface: $surface: ty) => {
```
i509VCB, line 155: This macro has limited usefullness if we have two user data implementations for something in a user's app.

danieldg: Hmm, it might be useful to change that to a list of types instead of a single type.

i509VCB (approved):

vberger (approved):

i509VCB:
It appears this has sat here for eons, time to merge I guess

PR #256, opened 2022-05-24
Update client registry handler API

danieldg:


src/registry.rs
```diff
@@ -152,109 +180,150 @@ pub enum BindError {
...
+    /// Binds a global, returning a new object associated with the global.
+    ///
+    /// This should not be used to bind globals that have multiple instances such as `wl_output`;
+    /// use [Self::bind_all] instead.
+    pub fn bind_one<I, D, U>(
+        &self,
         qh: &QueueHandle<D>,
-        name: u32,
-        version: u32,
+        version: std::ops::RangeInclusive<u32>,
```
i509VCB, line 238: Should we validate the range is valid (an assert)?

You could write `4..=1` and rust would not complain but it would bind the wrong version.

examples/image_viewer.rs
```diff
@@ -46,8 +43,9 @@ fn main() {
         windows: Vec::new(),
     };
 
-    event_queue.blocking_dispatch(&mut state).unwrap();
-    event_queue.blocking_dispatch(&mut state).unwrap();
+    while !state.registry_state.ready() {
+        event_queue.blocking_dispatch(&mut state).unwrap();
```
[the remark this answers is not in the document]
danieldg: `blocking_dispatch` is intended; there is already an active callback (which sets `ready = true` when complete)

src/registry.rs
danieldg: I don't see that as a big issue - it's much less likely than just passing in nonsense as the version range endpoints, which we also don't detect.

i509VCB (approved):

PR #257, opened 2022-05-25
Migrate Multipool to beta.1

snakedye:
This is a continuation of https://github.com/Smithay/client-toolkit/pull/220 .


src/shm/pool/multi.rs
```diff
@@ -0,0 +1,398 @@
...
+                } else {
+                    // If one of the overlapping buffers is busy, then no buffer can be returned because it could result in a data race.
+                    index = Err(PoolError::InUse);
+                }
+            } else if found_key {
+                break;
+            }
+            let mut size = buf_slot.size;
+            size += size % bpp;
+            size -= size % bpp;
```
danieldg, line 211: I don't think these statements do what you intend.  Assume `size = 256, bpp = 5`, then you get `257`, `255` as the new values for `size` and end up with a reserved buffer that is one byte short.  The same is true for `size=256, bpp = 3` (consider the Rgb888 format with aligned rows).

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,398 @@
...
+    ) -> Option<(usize, &wl_buffer::WlBuffer, &mut [u8])>
+    where
+        Q: PartialEq,
+        K: std::borrow::Borrow<Q>,
+    {
+        let len = self.inner.len();
+        let size = (stride * height) as usize;
+        let buf_slot =
+            self.buffer_list.iter_mut().find(|buf_slot| buf_slot.key.borrow().eq(key))?;
+        buf_slot.used = size;
```
danieldg, line 252: There's no check that the size requested for this key matches that used in `insert`, so it might overlap other buffers now.

danieldg, line 252: There's no check that the size requested for this key matches that used in `insert`, so it might overlap other buffers now.

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,398 @@
...
+
+    /// Calcule the offet and size of a buffer based on its stride.
+    fn offset(&self, mut offset: i32, width: i32, stride: i32, height: i32) -> (usize, usize) {
+        // bytes per pixel
+        let bpp = (stride as f32 / width as f32).ceil() as i32;
+        let size = stride * height;
+        // 5% padding.
+        offset += offset / 20;
+        offset += offset % bpp;
+        offset -= offset % bpp;
```
danieldg, line 340: Also a no-op.

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,398 @@
...
+    ) -> Result<usize, PoolError>
+    where
+        K: Borrow<Q>,
+        Q: PartialEq + ToOwned<Owned = K>,
+    {
+        let mut offset = 0;
+        let mut found_key = false;
+        let size = (stride * height) as usize;
+        let mut index = Err(PoolError::NotFound);
+        let bpp = (stride as f32 / width as f32).ceil() as usize;
```
danieldg, line 172: With a 5x5 ARGB8888 buffer whose rows are aligned to 8 bytes, you have `width = 5, stride = 24, bpp = 5` - likely not what is intended.

danieldg, line 172: With a 5x5 ARGB8888 buffer whose rows are aligned to 8 bytes, you have `width = 5, stride = 24, bpp = 5` - likely not what is intended.

src/shm/pool/multi.rs
danieldg, line 340: Same as above.

danieldg, line 340: Same as above.

src/shm/pool/multi.rs
```diff
@@ -0,0 +1,398 @@
...
+    where
+        Q: PartialEq,
+        K: std::borrow::Borrow<Q>,
+    {
+        if let Some((i, buf)) =
+            self.buffer_list.iter().enumerate().find(|(_, slot)| slot.key.borrow().eq(key))
+        {
+            let mut offset = buf.offset;
+            if let Some(buf) = self.buffer_list.remove(i).buffer {
+                buf.destroy();
```
danieldg, line 132: No check if the buffer is in use?  This risks making the surface undefined.

danieldg, line 132: No check if the buffer is in use?  This risks making the surface undefined.

src/shm/pool/multi.rs
snakedye: I just tried creating a 5x5 ARGB8888 buffer and the stride is 20. It's not aligned to 8 bytes from what I observed.

src/shm/pool/multi.rs
snakedye: >  Assume size = 256, bpp = 5

Is this scenario possible? The implementation assumes that there's no padding at the end of a stride. If the stride for each format is defined somewhere lmk.

src/shm/pool/multi.rs
danieldg: Correct, if your graphics library only requires 4-byte alignment then it will have a stride of 20.  However, if your graphics library uses 8- or 16-byte alignment for its buffers, it could be 24 or even 32.  This could be done in order to take advantage of 128-bit SSE instructions in rendering without needing to mask off bits at the end.

src/shm/pool/multi.rs
danieldg: The entire reason stride is specified is to permit padding.  The bit/byte width for each format is defined in the format enum itself (in the comments).

snakedye:
There are some formats I can't decipher. I just return a 0. It should probably panic but I think I can figure out their size.

src/shm/pool/format.rs
```diff
@@ -0,0 +1,117 @@
+//! Constants of bytes for pixel associated to wl_buffer's [`Format`]s.
+
+use wayland_client::protocol::wl_shm::Format;
+
+pub fn bytes_per_pixel(format: &Format) -> u32 {
```
i509VCB, line 5: Not immediately relevant for this pull request, this seems like data we could have in a crate that smithay (https://github.com/Smithay/smithay/pull/587) and sctk both could share.

Since drm formats are effectively the same as the wayland formats we could just have the macro generate functions for wl shm formats.

src/shm/pool/slot.rs
```diff
@@ -302,7 +302,9 @@ impl SlotPool {
         stride: i32,
         format: wl_shm::Format,
     ) -> Result<Buffer, CreateBufferError> {
-        let offset = slot.inner.offset as i32;
+        let mut offset = slot.inner.offset as i32;
+        offset += offset % 64;
+        offset -= offset % 64;
```
danieldg, line 307: This change just breaks things; it does not actually align anything.  The current alignment code is on line 259.

snakedye: I'm still not sure if I broke something.

src/shm/pool/slot.rs
```diff
@@ -255,8 +255,9 @@ impl SlotPool {
     }
 
     /// Create a new slot with the given size in bytes.
-    pub fn new_slot(&mut self, len: usize) -> io::Result<Slot> {
-        let len = (len + 15) & !15;
+    pub fn new_slot(&mut self, mut len: usize) -> io::Result<Slot> {
+        len += len % 64;
+        len -= len % 64;
```
danieldg, line 260: You want `let len = (len + 63) % 64;`

src/shm/pool/slot.rs
danieldg: The initial `x += x % y;` statement does not help with aligning `x`; that's what breaks things.  The second statement will correctly align values, but does so by rounding down - and at this point, that just ends up overlapping the prior buffer.

PR #258, opened 2022-05-26
Simplify PointerHandler

danieldg:
This exposes all the events in a pointer frame in one user callback instead of making them distinct function calls.  It also provides tracking of the active surface and pointer position for all events, which is useful for button press/release.

Axis (scroll) events now correctly report both discrete and absolute distances, instead of just the first of the two that was reported.

### DIFF
--- a/examples/simple_window.rs
+++ b/examples/simple_window.rs
@@ -9,7 +9,7 @@ use smithay_client_toolkit::{
     registry_handlers,
     seat::{
         keyboard::{KeyEvent, KeyboardHandler, Modifiers},
-        pointer::{PointerHandler, PointerScroll},
+        pointer::{PointerEvent, PointerEventKind, PointerHandler},
         Capability, SeatHandler, SeatState,
     },
     shell::xdg::{
@@ -54,7 +54,6 @@ fn main() {
         keyboard: None,
         keyboard_focus: false,
         pointer: None,
-        pointer_focus: false,
     };
 
     while !simple_window.registry_state.ready() {
@@ -111,7 +110,6 @@ struct SimpleWindow {
     keyboard: Option<wl_keyboard::WlKeyboard>,
     keyboard_focus: bool,
     pointer: Option<wl_pointer::WlPointer>,
-    pointer_focus: bool,
 }
 
 impl CompositorHandler for SimpleWindow {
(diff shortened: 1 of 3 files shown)


examples/simple_window.rs
```diff
@@ -331,94 +329,37 @@ impl KeyboardHandler for SimpleWindow {
...
+        for event in events {
+            // Ignore events for other surfaces
+            if Some(&event.surface) != self.window.as_ref().map(Window::wl_surface) {
+                continue;
             }
-
-            if let Some(horizontal) = scroll.axis(wl_pointer::Axis::HorizontalScroll) {
-                println!("\nH: {:?}", horizontal);
+            match event.kind {
+                Enter { .. } => {
```
i509VCB, line 346: Enter and leave should be separate from general events imo.

danieldg: One of the things I noticed when testing frames is that {Release, Leave} can be a single frame.  I don't know if that's a useful distinction that anyone will care about, but it is a reason to keep them together.  The {Leave, Enter} combination is called out in the wayland spec as a useful frame.

src/seat/pointer.rs
```diff
@@ -7,124 +7,87 @@ use wayland_client::{
...
-    /// The pos
-    fn pointer_focus(
-        &mut self,
-        conn: &Connection,
-        qh: &QueueHandle<Self>,
-        pointer: &wl_pointer::WlPointer,
-        surface: &wl_surface::WlSurface,
-        entered: (f64, f64),
+#[derive(Debug, Clone)]
+pub enum PointerEventKind {
```
i509VCB, line 50: Could this be copy?

Reason being that the pointer handler gives you a slice of events, which you need to clone or pass a vec to the user?

Maybe even a tinyvec since often there are very few events per frame (4 before leaking to the heap I guess)?

i509VCB: Plus the PointerEvent is trivial to copy.

danieldg: Using a smallvec for the events would make sense.  Making the kind `Copy` is possible, but `PointerEvent` can't be, so you lose most of the help that can provide.

i509VCB: oh yeah the surface...

i509VCB: I'd prefer tinyvec (no unsafe code) but smallvec is fine.

danieldg: I just chose `smallvec` because it's already a dependency.

vberger: We can migrate wayland-rs to tinyvec if it's a better alternative btw, I haven't looked into this yet.

danieldg: After looking at the API, we can't use tinyvec here because there is no `impl Default for WlSurface`.

i509VCB: > We can migrate wayland-rs to tinyvec if it's a better alternative btw, I haven't looked into this yet.

Yeah with the Default requirement in tinyvec it won't work here at least.

I believe you do end up with a slightly bigger struct in tinyvec because smallvec can use a union to save some space.

i509VCB (approved):

PR #259, opened 2022-05-26
bump: wl_output least supported version - 4

Shinyzenith:
This helps to populate name field when retrieving OutputInfo.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -191,8 +191,8 @@ impl crate::environment::MultiGlobalHandler<WlOutput> for OutputHandler {
         version: u32,
         _: DispatchData,
     ) {
-        // We currently support wl_output up to version 3
-        let version = std::cmp::min(version, 3);
+        // We currently support wl_output up to version 4
+        let version = std::cmp::min(version, 4);
         let output = registry.bind::<WlOutput>(version, id);
         let has_xdg;
         if let Some(xdg) = self.xdg_listener.as_ref().and_then(rc::Weak::upgrade) {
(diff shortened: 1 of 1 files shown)


src/output.rs
```diff
@@ -191,8 +191,8 @@ impl crate::environment::MultiGlobalHandler<WlOutput> for OutputHandler {
         version: u32,
         _: DispatchData,
     ) {
-        // We currently support wl_output up to version 3
-        let version = std::cmp::min(version, 3);
+        // We currently support wl_output up to version 4
+        let version = std::cmp::min(version, 4);
         let output = registry.bind::<WlOutput>(version, id);
         let has_xdg;
         if let Some(xdg) = self.xdg_listener.as_ref().and_then(rc::Weak::upgrade) {
```
vberger, line 198: If version 4 is available, is it still necessary to bind the xdg_output global?

Shinyzenith: Afaik smithay bound to xdg output just to get name and description events which are now fired by wloutput in v4. Should I unbind xdg output?

vberger: In that case, maybe adding a check on that if so that `has_xdg` is set to `false` if `version >= 4` ?

Shinyzenith: OK will get on it asap.

Shinyzenith: @vberger https://github.com/Smithay/client-toolkit/pull/259/commits/51a7bf9306a83a548f4a610280b6cdb6c3246074

vberger (approved):

PR #261, opened 2022-05-27
0.30: Add xdg_popup wrapper

danieldg:


src/shell/xdg/popup.rs
```diff
@@ -0,0 +1,260 @@
...
+}
+
+impl Drop for PopupInner {
+    fn drop(&mut self) {
+        self.xdg_popup.destroy();
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct PopupConfigure {
```
i509VCB, line 134: Could be non-exhaustive?

for context WindowConfigure and LayerSurfaceConfigure are non-exhaustive structs.

src/shell/xdg/popup.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// (x,y) relative to parent surface window geometry
+    pub position: (i32, i32),
+    pub width: i32,
+    pub height: i32,
+    pub serial: u32,
+    /// Only valid if [Self::is_reposition]
+    pub token: u32,
+    state: u32,
```
i509VCB, line 142: If this is internal data, this could probably just be an enum unless multiple states could be set at the same time.

src/shell/xdg/popup.rs
danieldg: The private `state` field already provides the same functionality as `#[non_exhaustive]`.

src/shell/xdg/popup.rs
danieldg: It's pulled from an `Atomic` and it can't be an enum when stored there.  I don't see much benefit from making it an enum just for half of the use, although it could have named constants.

src/shell/xdg/popup.rs
i509VCB: Sure, but I like to be explicit (just in case we end up removing the state field at a later date)

src/shell/xdg/inner.rs
```diff
@@ -10,7 +10,7 @@ where
     D: Dispatch<xdg_wm_base::XdgWmBase, ()> + XdgShellHandler + ProvidesRegistryState + 'static,
 {
     fn ready(data: &mut D, _conn: &Connection, qh: &QueueHandle<D>) {
-        data.xdg_shell_state().xdg_wm_base = data.registry().bind_one(qh, 1..=3, ()).into();
+        data.xdg_shell_state().xdg_wm_base = data.registry().bind_one(qh, 1..=4, ()).into();
```
i509VCB, line 13: xdg_toplevel::configure_bounds was introduced in version 4.

The toplevel dispatch implementation needs to handle the new event.

danieldg: Huh, I guess I thought that was already wired up because `WindowConfigure::suggested_bounds` exists.

src/shell/xdg/mod.rs
```diff
@@ -60,6 +61,49 @@ impl XdgShellState {
...
+}
+
+/// A trivial wrapper for an [`xdg_positioner::XdgPositioner`].
+///
+/// This wrapper calls [`destroy`][xdg_positioner::XdgPositioner::destroy] on the contained
+/// positioner when it is dropped.
+#[derive(Debug)]
+pub struct XdgPositioner(xdg_positioner::XdgPositioner);
+
+impl std::ops::Deref for XdgPositioner {
```
i509VCB, line 82: This is abusing Deref, this isn't a smart pointer.

danieldg: The same could be said of `std::lazy::Lazy` or `Cow::Owned`.  Requiring another function to access the contained value just clutters up user code for no particular benefit.

danieldg: Patch removed, will be part of #262 (modified to use that trait).

i509VCB (approved):

PR #262, opened 2022-06-01
Add ProvidesBoundGlobal trait

danieldg:
This allows decoupling users of globals from the `*State` types that bind them.

### DIFF
--- a/src/globals.rs
+++ b/src/globals.rs
@@ -45,3 +45,9 @@ pub trait ProvidesBoundGlobal<I: Proxy, const API_COMPAT_VERSION: u32> {
         }
     }
 }
+
+/// A struct used as the UserData field for globals bound by SCTK.
+///
+/// This is used instead of `()` to allow multiple `Dispatch` impls on the same object.
+#[derive(Debug)]
+pub struct GlobalData(pub(crate) ());
--- a/src/globals.rs
+++ b/src/globals.rs
@@ -1,0 +1,47 @@
+use crate::error::GlobalError;
+use wayland_client::Proxy;
+
+/// A trait implemented by types that provide access to capability globals.
+///
+/// The returned global must be fully compatible with the provided `API_COMPAT_VERSION` generic
+/// argument.  For example:
+///
+/// - A global that binds to `wl_compositor` with maximum version 4 could implement
+/// `ProvidesBoundGlobal<WlCompositor, 4>`, `ProvidesBoundGlobal<WlCompositor, 3>`,
+/// `ProvidesBoundGlobal<WlCompositor, 2>`, and `ProvidesBoundGlobal<WlCompositor, 1>` because
+/// versions 2-4 only add additional requests to the `wl_surface` API.
+/// - A global that binds to `wl_compositor` with maximum version 5 may only implement
+/// `ProvidesBoundGlobal<WlCompositor, 5>` because version 5 makes using `wl_surface::attach` with
+/// a nonzero offset a protocol error.  A caller who is only aware of the version 4 API risks
+/// causing these protocol errors if it uses surfaces created by such a global.
+///
+/// Changes that cause compatibility breaks include:
+///
+/// - Adding a new event to the global or to any object created by the global.
+/// - Adding a new requirement to an existing request.
+///
+/// The resulting global may have a version lower than `API_COMPAT_VERSION` if, at runtime, the
+/// compositor does not support the new version.  Clients should either be prepared to handle
+/// earlier versions of the protocol or use [`ProvidesBoundGlobal::with_min_version`] to produce an
+/// error in this case.
+///
+/// It is permitted to implement `ProvidesBoundGlobal` for versions that are higher than the
+/// maximum version you bind.  When rustc gains the ability to constrain const parameters with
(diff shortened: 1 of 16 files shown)


src/shm/mod.rs
```diff
@@ -49,6 +50,12 @@ impl ShmState {
     }
 }
 
+impl ProvidesBoundGlobal<wl_shm::WlShm, 1> for ShmState {
+    fn bound_global(&self) -> Result<wl_shm::WlShm, GlobalError> {
+        self.wl_shm().cloned()
```
i509VCB, line 55: I think you'll need `.map(Clone::clone)` or something like it.

i509VCB (commented):
More things could be put under this system as well. xdg shell and layer shell come to mind.

src/registry.rs
```diff
@@ -502,6 +502,53 @@ impl<I: Proxy> GlobalProxy<I> {
     }
 }
 
+#[derive(Debug)]
+pub struct SimpleGlobal<I, const MAX_VERSION: u32> {
```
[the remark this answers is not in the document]
danieldg: Yes, the goal of `SimpleGlobal` is to handle the most common case (you support all versions up to the last time you read the protocol docs).  The `with_min_version` can also be used to check the minimum version at runtime.

src/shm/mod.rs
danieldg: Yes, if we plan to leave the minimum rustc version as is.

i509VCB: This was stablized in 1.59, I believe we could jump to 1.60 or 1.61.

Winit which depends on the current sctk uses latest rust version.

src/registry.rs
```diff
@@ -502,6 +502,67 @@ impl<I: Proxy> GlobalProxy<I> {
...
+    I: Proxy,
+{
+    fn event(_: &mut D, _: &I, _: <I as Proxy>::Event, _: &(), _: &Connection, _: &QueueHandle<D>) {
+        unreachable!("SimpleGlobal is not suitable for {} which has events", I::interface().name);
+    }
+}
+
+#[macro_export]
+macro_rules! delegate_simple {
+    ($ty:ty, $iface:ty) => {
```
i509VCB, line 554: I think we should eliminate this first pattern. The user should know what version is being bound.

src/globals.rs
```diff
@@ -0,0 +1,47 @@
...
+/// argument.  For example:
+///
+/// - A global that binds to `wl_compositor` with maximum version 4 could implement
+/// `ProvidesBoundGlobal<WlCompositor, 4>`, `ProvidesBoundGlobal<WlCompositor, 3>`,
+/// `ProvidesBoundGlobal<WlCompositor, 2>`, and `ProvidesBoundGlobal<WlCompositor, 1>` because
+/// versions 2-4 only add additional requests to the `wl_surface` API.
+/// - A global that binds to `wl_compositor` with maximum version 5 may only implement
+/// `ProvidesBoundGlobal<WlCompositor, 5>` because version 5 makes using `wl_surface::attach` with
+/// a nonzero offset a protocol error.  A caller who is only aware of the version 4 API risks
+/// causing these protocol errors if it uses surfaces created by such a global.
```
i509VCB, line 16: Cc @vberger

vberger:
The general principle looks pretty good to me! Thanks a lot for taking the time to piece this together.

src/globals.rs
vberger: Sounds like a reasonable design principle to me, only provide multiple impl if there is no forward-compatibility hazard for the downstream user.

vberger: Hmm, I was under the impression on that doc test that it is ok to implement `ProvidesBoundGlobal` on `WlCompositor` from version 1 to 4 because essentially, if downstream requests version 1 and you provide version 4 instead, nothing will break in any way.

If providing `ProvidesBoundGlobal<WlCompositor, 2>` means you promise to not provide a high version than 2 even if they would be compatible, that means you cannot do early binding of the global, because you don't know which version the user will request.

danieldg: That's mostly correct. Providing `ProvidesBoundGlobal<WlCompositor, 2>` means that one of the following is true:

- You are providing a global whose maximum version is `<= 2`.  This could happen because the compositor only supports version 1, or it could be because you only bind version 1 even if version 2 (or 4) is available.
- You are providing a global which is API-compatible with version `2`. This requires that you look at the changes and make sure nothing will break.

This definition permits current code to implement `ProvidesBoundGlobal<WlCompositor, 6>` even if version 6 has not yet been defined, because code that is capable of supporting version 6 will be able to use the global that the current code binds - they might be versions 1-5, all of which someone requesting version 6 needs to be prepared to handle in case the compositor itself only supports that version.  This means that you can release a crate that uses newer features without waiting on an upgrade of both SCTK and wayland-protocols, but which will be able to use them once SCTK has updated.

danieldg:
Before I call this done, I think it makes sense to migrate more methods to use it - like change `XdgShellState::create_surface` to `XdgSurface::new(wm: &impl ProvidesBoundGlobal)`.

i509VCB (approved):
This all looks fine to me

vberger (approved):

PR #263, opened 2022-06-17
Update calloop to 0.10 + minor deps

vberger:


Cargo.toml
```diff
@@ -13,15 +13,15 @@ readme = "README.md"
 
 [dependencies]
 bitflags = "1.0"
-nix = "0.22"
+nix = "0.24"
```
i509VCB, line 16: Do we want minimal features?

src/seat/keyboard/mod.rs
```diff
@@ -12,18 +12,20 @@
...
 
+#[cfg(feature = "calloop")]
+use calloop::{timer::Timer, RegistrationToken};
 #[cfg(feature = "calloop")]
 use std::num::NonZeroU32;
 #[cfg(feature = "calloop")]
 use std::time::Duration;
 use std::{
-    cell::RefCell,
+    cell::{Cell, RefCell},
```
i509VCB, line 22: We could probably do something about the unusued imports?

Although like you said on Matrix, this really isn't important considering this code is going to be replaced soon.

i509VCB (approved):

Cargo.toml
vberger: Ah yeah, good point

PR #266, opened 2022-06-23
Buffer pool

i509VCB:
Calling this a pool is misleading to be fair. Could restructure directories to represent that case.

The buffer "pool" is a fixed size shm pool which backs a single wl_buffer.


i509VCB:
Effectively dead, if the demand is there I'll write a new pull request.

PR #269, opened 2022-07-06
Make GlobalData fully public

snakedye:
People that don't commit fully to the SCTK can use GlobalData to satisfy trait implementations necessary to use other SCTK types.


i509VCB (approved):
Just make sure to fix the lint and we can merge

vberger (approved):

PR #279, opened 2022-07-19
Simplify PopupInner by using Arc::new_cyclic

danieldg:
This saves a few bytes and is clearer about how the data is initialized.
It could be cleaner if std allowed the initialization closure to fail.


src/shell/xdg/popup.rs
```diff
@@ -88,42 +87,53 @@ impl Popup {
...
+            Some(PopupInner {
+                surface,
+                xdg_popup,
+                pending_position: (AtomicI32::new(0), AtomicI32::new(0)),
+                pending_dimensions: (AtomicI32::new(-1), AtomicI32::new(-1)),
+                pending_token: AtomicU32::new(0),
+                configure_state: AtomicU32::new(PopupConfigure::STATE_NEW),
+            })
+        });
+        debug_assert!(inner.is_some() || err.is_err());
```
i509VCB, line 115: Personally I think debug_assert is too weak. If it does return Err, then there is probably state corruption.

danieldg: This assert is something that can never happen (with the current code) - it's mainly there to help document the assumption, and doesn't need to be there in production.

vberger: I agree, but could you add a comment explaining that then?

vberger (approved):

i509VCB (approved):

PR #280, opened 2022-07-19
update some configure docs for window and layer surfaces

i509VCB:
This also adds some helpers for WindowConfigure to test if some state is set.

### DIFF
--- a/src/shell/layer/mod.rs
+++ b/src/shell/layer/mod.rs
@@ -52,21 +52,19 @@ impl LayerState {
     }
 }
 
+/// Handler for operations on a [`LayerSurface`]
 pub trait LayerHandler: Sized {
     fn layer_state(&mut self) -> &mut LayerState;
 
-    /// Called when the surface will no longer be shown.
+    /// The layer surface has been closed.
     ///
-    /// This may occur as a result of the output the layer is placed on being destroyed or the user has caused
-    /// the layer to be removed.
-    ///
-    /// You should drop the layer you have when this event is received.
+    /// When this requested is called, the layer surface is no longer shown and all handles of the [`LayerSurface`]
+    /// should be dropped.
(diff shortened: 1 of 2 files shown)


src/shell/layer/mod.rs
```diff
@@ -348,19 +346,19 @@ bitflags! {
...
     ///
-    /// The size is a hint, meaning the client is free to ignore the new size (if the client does not resize),
-    /// pick a smaller size to satisfy aspect ratio or resize in steps. If you pick a small size and the
-    /// surface is anchored to two opposite anchors, then surface will be centered on the axis.
+    /// The size is a hint, meaning the new size can be ignored. A smaller size could be picked to satisfy
+    /// some aspect ratio or resize in steps. If the size is smaller than suggested and the layer surface is
+    /// anchored to two opposite anchors then the layer surface will be centered on that axis.
     ///
-    /// If either the width or height is 0, the compositor may choose any size for that specific width or height.
+    /// If either the width or height is 0, any size dimensions may be used.
```
vberger, line 361: Something is missing here, the intent is that is the width is 0, you may choose any width you want, and same for the height.

MaxVerevkin: Would using `Option<NonZeroU32>` instead of `u32` help to show the intent? Also this would prevent some mistakes. By the way, having two named fields instead of a tuple might help too.

i509VCB: I'd like to be consistent across shells regarding the fields if possible.

I plan to add a `SurfaceConstraints` type in a later pull request. This type would provide some helpers to correctly set the size of the surface (considering things such as max width/height, configure bounds (applicable in xdg shell) and scale (I think scale would involve a way to scale some constraints.

vberger (approved):

PR #281, opened 2022-07-19
output: fix how remove_global is handled

MaxVerevkin:
Currently `remove_global` `.release()`s every output but the removed one.

---

By the way, wouldn't it make more sense to first run the callback and then call `.release()` and remove it from the list of outputs? Because currently, as far as I am aware, there is no way to get the output name/id from `output_destroyed` callback.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -636,27 +636,21 @@ where
         interface: &str,
     ) {
         if interface == "wl_output" {
-            let mut destroyed = vec![];
+            let output = data
+                .output_state()
+                .outputs
+                .iter()
+                .position(|o| o.name == name)
+                .expect("Removed non-existing output");
 
-            data.output_state().outputs.retain(|inner| {
-                let destroy = inner.name != name;
+            let wl_output = data.output_state().outputs[output].wl_output.clone();
+            data.output_destroyed(conn, qh, wl_output);
 
-                if destroy {
-                    if let Some(xdg_output) = &inner.xdg_output {
-                        xdg_output.destroy();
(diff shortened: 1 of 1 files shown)


src/output.rs
```diff
@@ -639,7 +639,7 @@ where
             let mut destroyed = vec![];
```
MaxVerevkin, line 639: I think this should be an `Option`

i509VCB: Hmm yeah that is a bit odd, only one surface can ever be destroyed at a time.

Instead of an `Option`, could we call `data.output_destroyed(conn, qh, output);` within the `retain` closure?

MaxVerevkin: No, there are borrow errors because `output_destroyed` takes `&mut self`.

MaxVerevkin: But it can be written without `retain`:

```rust
        if interface == "wl_output" {
            let output = data
                .output_state()
                .outputs
                .iter()
                .position(|o| o.name == name)
                .expect("Removed non-existing output");

            let wl_output = data.output_state().outputs[output].wl_output.clone();
            data.output_destroyed(conn, qh, wl_output);

            let output = data.output_state().outputs.remove(output);
            if let Some(xdg_output) = &output.xdg_output {
                xdg_output.destroy();
            }
            output.wl_output.release();
        }
```

i509VCB: Yeah that would be better.

vberger (approved):

i509VCB (approved):

PR #282, opened 2022-07-21
fix image_viewer example

MaxVerevkin:
Before 

https://user-images.githubusercontent.com/34583604/180189421-e6dbf496-e0e9-492c-8f28-86e43c1d07a3.mp4

After

https://user-images.githubusercontent.com/34583604/180189576-4fc906c2-639e-4320-b4ab-57292906ace9.mp4


snakedye:
I think what's actually inefficient is allocating a new Vec for the resized image. It would probably be better to use tiny-skia for this.

PR #284, opened 2022-07-23
example: image_viewporter

danieldg:
This demonstrates SimpleGlobal for a useful global.

### DIFF
--- a/examples/image_viewporter.rs
+++ b/examples/image_viewporter.rs
@@ -1,0 +1,327 @@
+use std::env;
+
+use smithay_client_toolkit::{
+    compositor::{CompositorHandler, CompositorState},
+    delegate_compositor, delegate_output, delegate_registry, delegate_shm, delegate_simple,
+    delegate_xdg_shell, delegate_xdg_window,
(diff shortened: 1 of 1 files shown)


examples/image_viewporter.rs
```diff
@@ -0,0 +1,327 @@
...
+    registry_handlers!(CompositorState, OutputState, ShmState, XdgShellState, XdgWindowState, SimpleGlobal<WpViewporter, 1>);
+}
+
+impl AsMut<SimpleGlobal<WpViewporter, 1>> for State {
+    fn as_mut(&mut self) -> &mut SimpleGlobal<WpViewporter, 1> {
+        &mut self.viewporter
+    }
+}
+
+impl Dispatch<WpViewport, ()> for State {
```
i509VCB, line 310: Is there not a delegate macro simple global provides for this?

danieldg: This is for `WpViewport`; the delegate is use (and can only be used) for the `WpViewporter` (which is the global).

i509VCB (approved):

PR #285, opened 2022-07-23
Expose the list of globals in RegistryState

### LINKED ISSUE #283: allow checking if a global is available with registry machinery
At the moment the registry helpers only allow binding a global and not querying if a global is available.

The use case for this is smithay's wayland backend. If `zwp_linux_dmabuf_v1` version 4 is not available, I need to check that wl_drm and zwp_linux_dmabuf_v1 version 3 are both available and would like to avoid binding one of those globals to realize the other is not available.

cc @danieldg

danieldg:
Fixes #283.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -187,11 +187,20 @@ pub struct RegistryState {
     ready: bool,
 }
 
-#[derive(Debug)]
-struct Global {
+#[derive(Debug, Clone, Eq, Hash, PartialEq, Ord, PartialOrd)]
+pub struct Global {
+    /// The numeric name for this global, usable in [RegistryState::bind_specific].
+    pub name: u32,
+    /// The maximum version supported by the compositor for this global.
+    pub version: u32,
     interface: String,
-    version: u32,
-    name: u32,
+}
+
+impl Global {
+    /// The interface name for this global.
+    pub fn interface(&self) -> &str {
+        &self.interface
+    }
 }
 
 impl RegistryState {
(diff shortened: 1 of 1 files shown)


src/registry.rs
```diff
@@ -231,6 +240,26 @@ impl RegistryState {
...
     }
 
+    /// Returns an iterator over all globals.
+    ///
+    /// This list may be incomplete if [Self::ready] returns false, and may change later if the
+    /// compositor adds or removes globals after initial enumeration.
+    ///
+    /// No guarantees are provided about the ordering of the globals in this iterator.
+    pub fn globals(&self) -> impl Iterator<Item = Global> + '_ {
+        self.globals.iter().cloned()
```
vberger, line 250: Why force the `.cloned()`, instead of returning `impl Iterator<Item = &Global>` ?

danieldg: I think I out-thought myself trying to plan for future changes to a Map<String, (version, name)> - but you could just as easily have a Set<Global> for that.

vberger (approved):

src/registry.rs
```diff
@@ -187,11 +187,20 @@ pub struct RegistryState {
...
 
-#[derive(Debug)]
-struct Global {
+#[derive(Debug, Clone, Eq, Hash, PartialEq, Ord, PartialOrd)]
+pub struct Global {
+    /// The numeric name for this global, usable in [RegistryState::bind_specific].
+    pub name: u32,
+    /// The maximum version supported by the compositor for this global.
+    pub version: u32,
     interface: String,
```
i509VCB, line 196: Seems a bit odd to have two public fields but then one only accessible via a function

i509VCB (approved):

src/registry.rs
danieldg: I did that to allow switching to `Box<str>` instead of `String` later, if desired.

PR #286, opened 2022-07-28
Rework layer surface and window around cyclic user data

i509VCB:
cc @danieldg


danieldg:
Looks good to me.  It may be useful to include changes from https://github.com/Smithay/wayland-rs/pull/531 to get rid of that extra `Option`.

i509VCB:
I'll integrate the changes from #531 when that is merged

vberger (approved):

PR #287, opened 2022-08-15
generic delegate macros

wash2:
This PR adds generic delegate macros to sctk. See https://github.com/wash2/client-toolkit/blob/master/examples/generic_list_seats.rs for an example.

### DIFF
--- a/examples/generic_list_seats.rs
+++ b/examples/generic_list_seats.rs
@@ -1,0 +1,95 @@
+use smithay_client_toolkit::{
+    delegate_registry, delegate_seat,
+    registry::{ProvidesRegistryState, RegistryState},
+    registry_handlers,
+    seat::{Capability, SeatHandler, SeatState},
+};
+use wayland_client::{protocol::wl_seat, Connection, QueueHandle};
+
+fn main() {
+    env_logger::init();
+
+    let conn = Connection::connect_to_env().unwrap();
+
+    let mut event_queue = conn.new_event_queue();
+    let qh = event_queue.handle();
+
+    let mut list_seats = ListSeats {
+        registry_state: RegistryState::new(&conn, &qh),
+        seat_state: SeatState::new(),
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -23,18 +23,23 @@ lazy_static = "1.0"
...
+wayland-protocols-wlr = { git="https://github.com/smithay/wayland-rs", features = ["client"] }
+wayland-cursor = { git="https://github.com/smithay/wayland-rs" }
 # Explicit dependency until release
 xkbcommon = { git = "https://github.com/rust-x-bindings/xkbcommon-rs", optional = true, features = ["wayland"] }
 calloop = { version = "0.10.0", optional = true }
 
 [features]
-default = ["calloop", "xkbcommon"]
+default = ["calloop", "xkbcommon", "system"]
+system = ["wayland-backend/client_system"]
```
vberger, line 42: Why add this feature, and why make it default?

Also this looks unrelated to this PR and should be discussed in a separate issue/PR imo.

wash2: I've been using as_ptr() for a while and recently needed to add this feature to make the method available again. I figured all features were enabled by default, so I enabled it by default. I can move it to a different PR though.

vberger: Mostly the encouraged method to use `as_ptr()` for a project is by itself depending directly on `wayland-backend` to activate the feature, as this method is part of `wayland-backend` API. So I don't see any reason to add that feature to SCTK, given it has 0 impact on the crate.

wash2: ok that change has been removed

wash2:
I've updated to resolve merge conflicts, but beta.9 still seems to break the examples for me, so I've left the git dependencies.

wash2:
I've updated to use the latest beta and squashed my changes.

i509VCB (approved):

PR #289, opened 2022-08-17
Don't require SeatHandler for KeyboardHandler

WhyNotHugo:
When using just wayland_client + KeyboardHandler, it is unnecessary for
the application's State to implement SeatHandler. It is, however,
necessary to implement KeyboardHandler when implementing SeatHandler.


WhyNotHugo:
I don't think the CI failure is due to my changes.

i509VCB (approved):

PR #291, opened 2022-08-20
Updates for infallible creation methods

danieldg:
This is the SCTK side of https://github.com/Smithay/wayland-rs/pull/531 and https://github.com/Smithay/wayland-rs/pull/530 - it moves failures out of the `Arc::new_cyclic` closure, removing the need for the `Option` there.

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -183,20 +183,36 @@ macro_rules! delegate_compositor {
     ($ty: ty) => {
         $crate::reexports::client::delegate_dispatch!($ty:
             [
-                $crate::reexports::client::protocol::wl_compositor::WlCompositor: $crate::globals::GlobalData,
-                $crate::reexports::client::protocol::wl_surface::WlSurface: $crate::compositor::SurfaceData,
-                $crate::reexports::client::protocol::wl_callback::WlCallback: $crate::reexports::client::protocol::wl_surface::WlSurface,
+                $crate::reexports::client::protocol::wl_compositor::WlCompositor: $crate::globals::GlobalData
+            ] => $crate::compositor::CompositorState
+        );
+        $crate::reexports::client::delegate_dispatch!($ty:
+            [
+                $crate::reexports::client::protocol::wl_surface::WlSurface: $crate::compositor::SurfaceData
+            ] => $crate::compositor::CompositorState
+        );
+        $crate::reexports::client::delegate_dispatch!($ty:
+            [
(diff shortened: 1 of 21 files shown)


src/compositor.rs
```diff
@@ -85,7 +85,7 @@ impl CompositorState {
     {
         let compositor = self.wl_compositor.get()?;
 
-        let surface = compositor.create_surface(qh, data)?;
+        let surface = compositor.create_surface(qh, data);
```
i509VCB, line 88: Does GlobalError still need an `InvalidId` variant?

i509VCB (changes requested):

src/compositor.rs
danieldg: We can probably remove all of the `InvalidId` variants after this patch, and maybe also some other errors (shm create_buffer can be infallible now).  That might be useful to add in this series.

src/shell/layer/mod.rs
```diff
@@ -127,38 +127,24 @@ impl LayerSurfaceBuilder {
     {
         // The layer is required in ext-layer-shell-v1 but is not part of the factory request. So the param
         // will stay for ext-layer-shell-v1 support.
+        let layer_shell = shell.bound_global()?;
+        let freeze = qh.freeze();
```
i509VCB, line 131: Let's explain why we need to freeze the queue during construction.

src/shell/xdg/popup.rs
```diff
@@ -80,46 +78,45 @@ impl Popup {
...
         });
-        // This assert checks that err was set properly above; it should be impossible to trigger,
-        // so it's not a run-time assert.
-        debug_assert!(inner.is_some() || err.is_err());
-        err.map(|()| Popup { inner })
+        drop(freeze);
+        Ok(Popup { inner })
     }
 
     fn inner(&self) -> &PopupInner {
```
i509VCB, line 118: This function is effectively redundant now.

i509VCB (approved):
This looks good, need to get CI to be green.

src/shell/xdg/popup.rs
danieldg: Which function? `from_surface` is not redundant with `new`.

i509VCB: The `inner` functions

danieldg: Ah. The github quote function makes this very unclear: it shows lines 117-121, and your comment is about line 121-123.

PR #295, opened 2022-09-04
Update to 0.30.0-beta.9 release

hw0lff:
I bumped all wayland-rs versions to beta.9 and fixed the compile errors by removing error propagation where it was not needed anymore.


i509VCB:
I think #291 already covers this?

hw0lff:
Yes, it does.

PR #299, opened 2022-09-19
Add `update_keymap` method that's called when server provides keymap

ids1024:
This is meant to provide a way to address
https://github.com/pop-os/xdg-shell-wrapper/issues/8, in combination with https://github.com/Smithay/smithay/pull/750.

This is currently potentially unsound due to the
`unsafe impl Send for KeyboardData {}`. `xdg::Keymap` uses non-thread-safe ref-counting, but sctk assumes it is only used from one thread. This is difficult to address.

### DIFF
--- a/src/seat/keyboard/mod.rs
+++ b/src/seat/keyboard/mod.rs
@@ -94,6 +94,17 @@ impl SeatState {
     }
 }
 
+/// Wrapper around a libxkbcommon keymap
+#[allow(missing_debug_implementations)]
+pub struct Keymap<'a>(&'a xkb::Keymap);
+
+impl<'a> Keymap<'a> {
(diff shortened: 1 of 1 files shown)


i509VCB (approved):
Looks fine to me. Is this still a draft for a reason?

ids1024:
The API isn't great here, but I don't know if anything better is really possible. Otherwise this should work fine. (While I think https://github.com/Smithay/smithay/pull/750 isn't quite right.)

i509VCB:
I don't see anything else that would need to be done, @ids1024 are you fine with this being merged or do you have some new idea to handle this.

ids1024:
Merging it seems good to me, given it's useful to support and I can't really think of a better way to do so.

PR #301, opened 2022-10-04
feat: data device handling

wash2:
This PR adds data device handling and an example demonstrating its use. 
I've also copied over the `ReadPipe` implementation from the previous version of sctk, because it seems useful.
I haven't added very many abstractions, aside from the data sources, one for copy paste & another for drag and drop, to encourage that they are used correctly. 

The example will read from the selection whenever it changes and can accept DnD. 
If the user presses left click, it will start a DnD.
If the user presses "c" it will set the selection for copy paste.


src/data_device/data_source.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    /// Request to send data from the client.
+    /// Send the data, then close the fd.
+    fn send(
+        &mut self,
+        conn: &Connection,
+        qh: &QueueHandle<Self>,
+        source: &WlDataSource,
+        mime: String,
+        fd: OwnedFd,
```
i509VCB, line 48: Sadly OwnedFd is not `#[must_use]` I will open an issue about that in std.

src/data_device/data_source.rs
```diff
@@ -0,0 +1,182 @@
...
+        &self
+    }
+}
+
+/// Handler trait for DataSource events.
+///
+/// The functions defined in this trait are called as DataSource events are received from the compositor.
+pub trait DataSourceHandler: Sized {
+    /// The accepted mime type from the destination, if any
+    fn accept_mime(
```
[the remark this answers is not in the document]
wash2: Ya, I figured my doc comments needed more explanation. I'll improve this one and revisit the rest of them as well.

src/data_device/mod.rs
```diff
@@ -0,0 +1,315 @@
...
+        readiness: calloop::Readiness,
+        token: calloop::Token,
+        mut callback: F,
+    ) -> std::io::Result<calloop::PostAction>
+    where
+        F: FnMut((), &mut fs::File),
+    {
+        self.file.process_events(readiness, token, |_, file| {
+            callback((), file);
+            Ok(calloop::PostAction::Continue)
```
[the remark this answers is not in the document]
wash2: I believe it is not meant to be a oneshot, but I think that was just an easy way for me to use it in the example. A user really probably shouldn't try to read all the data from the pipe at once, considering it might take a while.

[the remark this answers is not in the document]
i509VCB: If that is the case, then how would you know all the data has been copied if readiness could lie?

Cargo.toml
```diff
@@ -46,6 +46,8 @@ env_logger = "0.9"
 wgpu = "0.13.1"
 raw-window-handle = "0.4.3"
 pollster = "0.2.5"
+gfx-hal = "*"
```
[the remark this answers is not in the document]
wash2: oh uh I'm not sure why that is there, sorry

src/data_device/data_source.rs
```diff
@@ -0,0 +1,182 @@
...
+    ) {
+        match event {
+            wl_data_source::Event::Target { mime_type } => {
+                state.accept_mime(conn, qh, source, mime_type)
+            }
+            wl_data_source::Event::Send { mime_type, fd } => {
+                state.send(conn, qh, source, mime_type, fd);
+            }
+            wl_data_source::Event::Cancelled => {
+                source.destroy();
```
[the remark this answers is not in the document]
wash2: I don't think I understand what you mean. Do you mean destroying the source? I don't think there is any other necessary cleanup other than the data source, if that is what you mean.

src/data_device_manager/data_device.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    /// get the active dnd offer if it exists
+    fn dnd_offer(&self) -> Option<DndDataOffer> {
+        self.data_device_data().inner.lock().ok().and_then(|inner| inner.dnd_data_offer.clone())
+    }
+
+    /// get the active selection offer if it exists
+    fn selection_offer(&self) -> Option<WlDataOffer> {
+        self.data_device_data().inner.lock().ok().and_then(|inner| inner.selection.clone())
+    }
```
i509VCB, line 58: These two functions are override-able to  handle dnd and selection offers that an extension may use I assume

src/data_device_manager/data_device.rs
```diff
@@ -0,0 +1,218 @@
...
+    fn data_offer(
+        &mut self,
+        conn: &Connection,
+        qh: &QueueHandle<Self>,
+        data_device: &WlDataDevice,
+        offer: WlDataOffer,
+    );
+
+    /// The data device pointer has entered a surface at the provided location
+    #[allow(clippy::too_many_arguments)]
```
i509VCB, line 81: Is this lint dead?

src/data_device_manager/data_device.rs
```diff
@@ -0,0 +1,218 @@
...
+        data: &U,
+        conn: &Connection,
+        qh: &QueueHandle<D>,
+    ) {
+        let data = data.data_device_data();
+        let mut inner = data.inner.lock().unwrap();
+
+        match event {
+            wayland_client::protocol::wl_data_device::Event::DataOffer { id } => {
+                drop(inner);
```
i509VCB, line 130: We should note this drop is required where it's used in this dispatch impl to prevent Mutex deadlocks.

src/data_device_manager/data_source.rs
```diff
@@ -0,0 +1,202 @@
...
+#[derive(Debug, PartialEq, Eq)]
+pub struct CopyPasteSource {
+    pub(crate) inner: WlDataSource,
+    pub(crate) serial: Option<u32>,
+}
+
+impl CopyPasteSource {
+    /// set the selection
+    /// internally tracks the serial for when unset selection may be called
+    pub fn set_selection(&self, device: &WlDataDevice, serial: u32) {
```
i509VCB, line 122: I am not too sure what this is supposed to do from the documentation here.

wash2: This method is really just a wrapper that calls set_selection. After looking at it again, I think it shouldn't hold onto the serial so I removed that part and adjusted the docs.

src/data_device_manager/mod.rs
```diff
@@ -0,0 +1,333 @@
...
+        D: Dispatch<WlDataSource, DataSourceData> + 'static,
+        I: IntoIterator<Item = &'s str>,
+    {
+        let source = self.create_data_source_with_data(qh, Default::default());
+
+        for mime in mime_types {
+            source.offer(mime.to_string());
+        }
+        if let Some(dnd_actions) = dnd_actions {
+            source.set_actions(dnd_actions);
```
ids1024, line 94: Apparently `set_actions()` was added in version 3 of the protocol, while this code is binding version `1..=3`. So presumably this could result in a protocol error on a compositor implementing version 1 or 2.

ids1024:
`wl_data_offer::finish` was also added in version 3. I guess currently that's not called in the library code but is used directly in the example. Perhaps the library should offer some API that calls that if supported? (However that would work.)

I would be good if sctk could smooth out any differences between versions of this protocol and just have a high level API that won't break on an old protocol version, and guarantees you won't call a method that's only valid for drag-and-drop with a copy-paste source.

wash2:
> `wl_data_offer::finish` was also added in version 3. I guess currently that's not called in the library code but is used directly in the example. Perhaps the library should offer some API that calls that if supported? (However that would work.)
> 
> I would be good if sctk could smooth out any differences between versions of this protocol and just have a high level API that won't break on an old protocol version, and guarantees you won't call a method that's only valid for drag-and-drop with a copy-paste source.

Ok, I will work on some API improvements

src/data_device_manager/data_offer.rs
```diff
@@ -0,0 +1,579 @@
...
+        };
+    }
+
+    pub(crate) fn to_selection_offer(&self) {
+        let mut inner = self.inner.lock().unwrap();
+        match inner.deref_mut() {
+            DataDeviceOffer::Drag(o) => {
+                *inner = DataDeviceOffer::Selection(SelectionOffer {
+                    data_offer: o.data_offer.take(),
+                    mime_types: o.mime_types.drain(..).collect(),
```
ids1024, line 297: You can use `std::mem::take()` to replace the vec with an empty vec. Which should just move the pointer instead of creating a new allocation.

The pattern is a bit awkward to deal with in Rust, but perhaps it would be better to use `inner: Arc<Mutex<Option<DataDeviceInner>>>`? So you just `.take()` the inner value, and then consume the whole `DataDeviceOffer`.

src/data_device_manager/data_device.rs
```diff
@@ -0,0 +1,294 @@
...
+pub trait DataDeviceHandler: Sized {
+    /// Introduces a new data offer
+    /// ASHLEY left out because the data offer will be introduced to the user once the type is known
+    // fn data_offer(
+    //     &mut self,
+    //     conn: &Connection,
+    //     qh: &QueueHandle<Self>,
+    //     data_device: DataDevice,
+    //     offer: WlDataOffer,
+    // );
```
i509VCB, line 138: Mentioning the other traits here for maintainers would probable be fine.

src/data_device_manager/data_device.rs
```diff
@@ -0,0 +1,294 @@
...
+        self
+    }
+}
+
+/// Handler trait for DataDevice events.
+///
+/// The functions defined in this trait are called as DataDevice events are received from the compositor.
+pub trait DataDeviceHandler: Sized {
+    /// Introduces a new data offer
+    /// ASHLEY left out because the data offer will be introduced to the user once the type is known
```
i509VCB, line 131: These are doc comments and actually apply to `DataDeviceHandler::entered`

i509VCB (approved):
I don't have much more that I can find needing any changes.

PR #305, opened 2022-10-06
egl example

### LINKED ISSUE #304: EGL example using glutin 0.30
The new glutin is now independent of winit, so we can use it more easily for an example. This will probably depend on #303

i509VCB:
Resizing does not behave correctly on KDE at the moment.

Fixes #304


i509VCB:
This needs to be redone

PR #306, opened 2022-10-07
feat: pointer themes

wash2:
This is an update & refactor of the work started in https://github.com/Smithay/client-toolkit/pull/228 

I've added a working example for customizing the pointer.


wash2:
I've refactored a bit while merging so that the pointer theming is part of the seat & pointer instead of shm

examples/pointer_theme_window.rs
```diff
@@ -0,0 +1,491 @@
...
+        &mut self.shm_state
+    }
+}
+
+impl SimpleWindow {
+    pub fn draw(&mut self, conn: &Connection, qh: &QueueHandle<Self>) {
+        if let Some(window) = self.window.as_ref() {
+            if self.themed_pointer.is_none() && self.pointer.is_some() {
+                let ptr = self.pointer.clone().unwrap();
+                let ptr_data: &PointerData = ptr.data().unwrap();
```
i509VCB, line 380: Couldn't we return both the ThemedPointer instance and the WlPointer at the same time? Not very fun to have to poke into the user data all the time.

This seems to force a pointer when created to have a theme with it, which for most cases is fine. But sometimes it may be desired to not touch the cursor theme at all.

wash2: > Couldn't we return both the ThemedPointer instance and the WlPointer at the same time? Not very fun to have to poke into the user data all the time.

Ok, I see what you mean. I think I can improve that.

> This seems to force a pointer when created to have a theme with it, which for most cases is fine. But sometimes it may be desired to not touch the cursor theme at all.

If I understand your question correctly, you mean like in the case of an application which wants to set the cursor image themselves and doesn't want to use the ThemedPointer to set the cursor at all? `get_pointer` is available which just sets the ThemedPointer to be default, and I don't believe it will have an effect if it isn't used, but I think I could probably put it behind a feature so it's left out completely if the user doesn't want to use it.

examples/pointer_theme_window.rs
```diff
@@ -0,0 +1,483 @@
...
+    xdg_window_state: XdgWindowState,
+
+    exit: bool,
+    first_configure: bool,
+    pool: Option<SlotPool>,
+    width: u32,
+    height: u32,
+    shift: Option<u32>,
+    buffer: Option<Buffer>,
+    window: Option<Window>,
```
i509VCB, line 121: The simple window example probably has been modified to compensate for the registry changes, we can create the window before looping for hotpluggable globals, effectively removing the need for an Option around Buffer and window.

src/seat/pointer/mod.rs
```diff
@@ -341,3 +352,178 @@ where
...
+        /// for HiDPI outputs.
+        size: u32,
+    },
+
+    /// Use the system provided theme
+    ///
+    /// In this case SCTK will read the `XCURSOR_THEME` and
+    /// `XCURSOR_SIZE` environment variables to figure out the
+    /// theme to use.
+    System,
```
i509VCB, line 402: We could make this the `default` value. You'd need to implement Default by hand since the `#[default]` attribute was only stabilized recently (I think 1.64)

src/seat/mod.rs
```diff
@@ -102,6 +105,38 @@ impl SeatState {
...
+        theme: ThemeSpec,
+        scale: i32,
+    ) -> Result<(wl_pointer::WlPointer, ThemedPointer), SeatError>
+    where
+        D: Dispatch<wl_pointer::WlPointer, PointerData> + PointerHandler + 'static,
+    {
+        let wl_ptr = self.get_pointer_with_data(
+            qh,
+            seat,
+            PointerData { inner: Mutex::new(PointerDataInner { ..Default::default() }) },
```
i509VCB, line 126: This could just be `PointerData::default()`?

src/seat/mod.rs
```diff
@@ -102,6 +105,38 @@ impl SeatState {
         self.get_pointer_with_data(qh, seat, PointerData::default())
     }
 
+    /// Creates a pointer from a seat with the provided theme.
+    ///
+    /// ## Errors
+    ///
+    /// This will return [`SeatError::UnsupportedCapability`] if the seat does not support a pointer.
+    pub fn get_pointer_with_theme<D>(
```
i509VCB, line 113: Seems like we also need a `get_pointer_with_theme_and_data`?

These `with_xyz` functions don't scale very well...

examples/pointer_theme_window.rs
i509VCB: Yes

i509VCB (approved):
lgtm

PR #234, opened 2022-04-07
0.30: keyboard repeating source

i509VCB:

PR #307, opened 2022-10-07
feat: keyboard repetition

wash2:
This is a refactor of https://github.com/Smithay/client-toolkit/pull/234 . It uses the latest Calloop Timer, and I adapted the simple window to include its usage.

The only point that I am unsure about handling timer stop & start. It wants to remove itself from the event loop after TimeoutAction::Drop, or it disagrees with the channel post action, and they continually re-register, which seems to cause busy looping. I hope I've handled it well, as my current implementation here seems to avoid the repeated calls to processing,


src/seat/keyboard/mod.rs
```diff
@@ -568,10 +589,23 @@ where
...
+                                        let _ = repeat_sender.send(RepeatMessage::StopRepeat);
+                                    }
+                                }
                                 data.release_key(conn, qh, keyboard, serial, event);
                             }
 
                             wl_keyboard::KeyState::Pressed => {
+                                #[cfg(feature = "calloop")]
+                                {
+                                    if let Some(repeat_sender) = &udata.repeat_sender {
```
i509VCB, line 604: xkbcommon has keys which should not be repeated. I believe there is a way to check for those keys on the keymap?

i509VCB (commented):
Should the user data with the keyboard send a stop repeat event when dropped?

i509VCB (approved):
Looks fine to me.

@elinorbgr, any comments?

elinorbgr:
The general structure looks pretty good, but I think there is an edge case that is not taken into account: repeat involving multiple pressed keys. So there is no "standard" way to handle that (not all apps to the same thing), but here are two scenarios that describe the behavior of the currently-released SCTK:

Scenario 1:
- key `A` is pressed
  - the character `a` starts repeating
- key `B` is pressed
  - the repeating character is changed to `b`
- key `A` is released
  - character `b` continues repeating
- key `B` is released
  - repeating stops

Scenario 2:
- key `A` is pressed
  - the character `a` starts repeating
- key `shift` is pressed
  - the repeating character changes to `A`
- key `shift` is released
  - the repeated character changes to `a`
- key `A` is released
  - repeating stops

wash2:
@elinorbgr The most recent commit should handle the two scenarios of multiple pressed keys in the same way as you describe

src/seat/keyboard/repeat.rs
```diff
@@ -0,0 +1,242 @@
...
+        // Re-register the timer to start it again
+        if reregister {
+            return Ok(PostAction::Reregister);
+        }
+
+        let timer_pa = timer.process_events(readiness, token, |mut event, _| {
+            if self.disabled || key.is_none() {
+                // TODO How to pause the timer without dropping it?
+                // return TimeoutAction::ToDuration(Duration::from_millis(*delay_mut));
+                return TimeoutAction::Drop;
```
i509VCB, line 200: bringing attention to this for @elinorbgr

src/seat/keyboard/repeat.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    /// The key event should not have any time added, the repeating mechanism is responsible for that instead.
+    StartRepeat(KeyEvent),
+
+    RepeatInfo(RepeatInfo),
+}
+
+/// [`EventSource`] used to emit key repeat events.
+#[derive(Debug)]
+pub struct KeyRepeatSource {
```
i509VCB, line 31: Might want to mark this as `#[must_use]`

i509VCB (approved):
lgtm after the change I suggested

src/seat/keyboard/repeat.rs
elinorbgr: Dropping it and recreating a new one later is the intended usage on calloop, if that a problem?

elinorbgr (approved):

src/seat/keyboard/repeat.rs
wash2: Ya, I guess, the issue was that that caused the PostAction to be PostAction::Remove iirc. But I think the logic that I have at line 213 for selecting a PostAction  works. I was just unsure that I had done it the best way.

PR #311, opened 2022-10-17
fix: return Result on MultiPool.get_at

snakedye:


i509VCB (approved):
I guess lint fixing broke it.

PR #313, opened 2022-10-25
Make `RegistryState` depend on use of `registry_queue_init`

ids1024:
This prevents the need to create two registry objects in examples like `simple_window`, and eliminates the need to manually implement `Dispatch`.

This would perhaps make it harder if an application wants to handle custom globals, instead of just putting that in the `Dispatch` implementation, but if the `RegistryState` helper doesn't already provide an ergonomic way to do that, maybe the way the `ProvidesRegistryState` trait and `registry_handlers!` macro work isn't ideal?

Leaving as a draft because I think it might be good to remove `ready()` and make things like `OutputState::new()` also take a `&GlobalList`. But does this seem like a generally good approach?

### DIFF
--- a/examples/list_outputs.rs
+++ b/examples/list_outputs.rs
@@ -24,10 +24,10 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     // Initialize the registry handling so other parts of Smithay's client toolkit may bind
     // globals.
-    let registry_state = RegistryState::new(&globals, &conn, &qh);
+    let registry_state = RegistryState::new(&globals);
 
     // Initialize the delegate we will use for outputs.
-    let output_delegate = OutputState::new();
+    let output_delegate = OutputState::new(&globals, &qh);
 
     // Set up application state.
     //
@@ -35,15 +35,10 @@ fn main() -> Result<(), Box<dyn Error>> {
     // application is running.
     let mut list_outputs = ListOutputs { registry_state, output_state: output_delegate };
 
-    // Send requests to the server and block until we receive events from the server.
-    while !list_outputs.registry_state.ready() {
-        event_queue.blocking_dispatch(&mut list_outputs)?;
-    }
-
-    // We do this again here because the first time we let the delegates bind their globals.
+    // `OutputState::new()` binds the output globals found in `registry_queue_init()`.
     //
-    // After the delegates have bound their globals and created the necessary objects, we need to dispatch
-    // again so that events may be sent to the newly created objects.
+    // After the globals are bound, we need to dispatch again so that events may be sent to the newly
+    // created objects.
     event_queue.roundtrip(&mut list_outputs)?;
 
     // Now our outputs have been initialized with data, we may access what outputs exist and information about
--- a/examples/list_outputs.rs
+++ b/examples/list_outputs.rs
@@ -8,7 +8,7 @@ use smithay_client_toolkit::{
     registry::{ProvidesRegistryState, RegistryState},
     registry_handlers,
 };
-use wayland_client::{protocol::wl_output, Connection, QueueHandle};
+use wayland_client::{globals::registry_queue_init, protocol::wl_output, Connection, QueueHandle};
 
 fn main() -> Result<(), Box<dyn Error>> {
     // We initialize the logger for the purpose of debugging.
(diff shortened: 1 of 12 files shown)


ids1024:
For removing `ready()`:

* The `RegistryHandler` implementation for `SimpleGlobal` needs to be removed, since that only provides `ready()`. `SimpleGlobal::not_ready()` then becomes useless, I believe.
* `OutputState::new()` needs to be changed to take a `GlobalList`, `Connection`, and `QueueHandle`, and can do what `ready()` currently does.
* The same with `SeatState::new()`
  - This poses an issue, since `ready()` calls `new_seat`, but `SeatState::new()` can't call methods on the state object. It could be changed to only call that method for new seats that are dynamically added? Not sure what solution would be suitable otherwise.
* `RegistryState::bind_all()` couldn't be used in these `new()` functions (unless they accept a `RegistryState`?) So the logic there needs to be moved to a standalone function or duplicated.

i509VCB:
Thinking further over this, I think this solution is fine enough for now. Are there any things in particular you wanted to look into?

ids1024:
Implemented the changes suggested in my last comment, with the potential concerns mentioned there and in the commit.

Seems like a generally better API than previously. The way I copied `bind_all` and `bind_once` as standalone functions perhaps being the exception. Not sure what would be better there.

src/seat/mod.rs
```diff
@@ -354,26 +375,6 @@ impl<D> RegistryHandler<D> for SeatState
...
-                name: Arc::new(Mutex::new(None)),
-                id,
-            })
-            .expect("failed to bind global");
-
-        for seat in seats {
-            let data = seat.data::<SeatData>().unwrap().clone();
-
-            state.seat_state().seats.push(SeatInner { seat: seat.clone(), data });
-            state.new_seat(conn, qh, seat);
```
i509VCB, line 373: When we create the seat state, there is no notification initially that any new seats were created. Or should it be expected that the user would iterate over the initially available seats?

src/output.rs
```diff
@@ -130,8 +131,36 @@ impl fmt::Debug for ScaleWatcherHandle {
...
+    pub fn new<
+        D: Dispatch<wl_output::WlOutput, OutputData>
+            + Dispatch<zxdg_output_v1::ZxdgOutputV1, OutputData>
+            + Dispatch<zxdg_output_manager_v1::ZxdgOutputManagerV1, GlobalData>
+            + 'static,
+    >(
+        global_list: &GlobalList,
+        qh: &QueueHandle<D>,
+    ) -> OutputState {
+        let (outputs, xdg) = global_list.contents().with_list(|globals| {
```
i509VCB, line 143: Like seats, there is no notification via the handle trait that an outputs existed at the start. Same question but with outputs pretty much.

src/seat/mod.rs
ids1024: > Or should it be expected that the user would iterate over the initially available seats?

This doesn't seem unreasonable since it's the same as how the use would be expected to handle globals, if they used `GlobalState` to iterate over globals available initially and detect when any are added/removed.

Otherwise, it runs into the problem of how to defer calling that callback until the state is available.

src/output.rs
ids1024: The behavior here isn't really changed. `new_output()` isn't called until `done` is sent. So it still occurs basically as previously, I believe.

examples/list_outputs.rs
```diff
@@ -19,27 +19,22 @@ fn main() -> Result<(), Box<dyn Error>> {
...
     // This is where you will store your delegates and any data you wish to access/mutate while the
     // application is running.
     let mut list_outputs = ListOutputs { registry_state, output_state: output_delegate };
 
-    // Send requests to the server and block until we receive events from the server.
-    while !list_outputs.registry_state.ready() {
-        event_queue.blocking_dispatch(&mut list_outputs)?;
-    }
-
     // We do this again here because the first time we let the delegates bind their globals.
```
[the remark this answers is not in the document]
ids1024: Good point. It is a second roundtrip, still, after the one that occurs within `registry_queue_init()`, but the comment doesn't really make sense as is.

i509VCB (approved):
lgtm

PR #322, opened 2022-11-23
shell rework

### LINKED ISSUE #316: Ensure all shell implementations do not internally commit buffers under the hood.
See https://github.com/rust-windowing/winit/issues/2164

i509VCB:
Taking a hammer to the shell abstractions and reworking those a bit.

Fixes #316


wash2:
Is this pretty much ready? 

Would this also be a good time to make any changes to popup creation? Correct me if I'm wrong, but the [Popup::new](https://smithay.github.io/client-toolkit/smithay_client_toolkit/shell/xdg/popup/struct.Popup.html#method.new) method can't be used to create a popup which requests a grab, which I found a bit awkward. [Popup::from_surface](https://smithay.github.io/client-toolkit/smithay_client_toolkit/shell/xdg/popup/struct.Popup.html#method.from_surface) still worked for my purposes which was fine, but I figure it's worth a mention.

Also, you didn't change it here, but the generic_list_seats example doesn't seem to be working when i ran it because of an extra call to blocking_dispatch. If you want to update that or remove it, either way is fine with me :)

i509VCB:
Yes this is ready

wash2 (approved):

wash2 (approved):

PR #323, opened 2022-11-27
remove unnecessary Result in `CompositorState::create_surface`

MaxVerevkin:


HeavyRain266:
Could you explain how is it unecessary? I'm curious because before removal looked fine to me 🤔.

MaxVerevkin:
> Could you explain how is it unecessary?

`create_surface()` always returns `Ok(surface)`

wash2 (approved):
This looks good to me :)

PR #325, opened 2022-11-28
remove unnecessary code that lazily binds xdg-output-manager

MaxVerevkin:
Currently, xdg-output-manager is bound during OutputState creation unconditionally, so there is no need to check it.

Please correct me if I'm wrong.


wash2 (approved):
Ya, this doesn't seem to be necessary.

PR #326, opened 2022-11-29
output: remove some duplicating code and simplify wl_output::Event::Mode hangling

MaxVerevkin:


wash2 (approved):
Thanks for the cleanup here!

PR #327, opened 2022-11-29
output: pending_xdg is written is never read

MaxVerevkin:


src/output.rs
```diff
@@ -495,16 +487,6 @@ where
                 inner.current_info = Some(info.clone());
                 inner.pending_wl = false;
 
-                if inner
```
i509VCB, line 498: I believe this will still be needed for version 3 of xdg-output. Although our current interpretation of xdg-output might be wrong.

xdg_output::done states the following:
> This event is sent after all other properties of an xdg_output have been sent.

Now this doesn't talk about the properties of wl_output AND xdg_output, but only xdg_output. Don't worry about the technical issue I mentioned, that probably needs to be addressed elsewhere.

PR #329, opened 2022-12-15
WIP `ext-screencopy-v1`

ids1024:
Based on the latest version of https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/124.

May be useful to test the protocol, once there are any server implementations of the latest revision.


ids1024:
https://github.com/pop-os/cosmic-protocols/tree/main/client-toolkit has an implementation of this. I may open A PR later adding a version of that here, but this PR probably isn't that useful.

PR #331, opened 2022-12-22
output: Don't call `update_output` at start when xdg outputs are present

ids1024:
With `zxdg_output_v1::done` deprecated in version 3, both Smithay and Sway seem to send `wl_output::done` twice at the start here, once after populating `wl_output` parameters, and again after populating the xdg output. So `new_output()` is immediately followed by `update_output()`.

This defers calling `new_output()` until we've gotten at least one xdg output event, which seems to work. I can't think of a better way to do it.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -495,24 +495,17 @@ where
                 inner.current_info = Some(info.clone());
                 inner.pending_wl = false;
 
-                if inner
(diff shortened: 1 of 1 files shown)


wash2 (approved):
This looks good to me :+1:

PR #334, opened 2023-01-01
event_loop: remove the WaylandSource

kchibisov:
The `calloop`'s WaylandSource integration is a part of the wayland-client crate.

--

It won't build without `log` until the https://github.com/Smithay/wayland-rs/pull/594


i509VCB:
This looks good, pending the wayland-rs pull request.

kchibisov:
Hm, given that sctk forces log(it always has it in a dep), shouldn't it be changed to provide it via feature as well?

i509VCB (approved):
we will need another wayland-rs release in order to remove the todo in the cargo.toml

kchibisov:
Yeah, even a patch release would be nice, given that the build is broken entirely right now and that wasn't noticed.

PR #335, opened 2023-01-02
output: add `OutputData::with_output_info`

kchibisov:
The `OutputInfo` is already stored in the data, so this commit makes access to it.

--

I think it's a leftover that it wasn't ported.

### DIFF
--- a/src/output.rs
+++ b/src/output.rs
@@ -240,10 +240,19 @@ impl OutputData {
         OutputData(Arc::new(Mutex::new(OutputInfo::new(name))))
     }
 
+    /// Get the output scale factor.
     pub fn scale_factor(&self) -> i32 {
         let guard = self.0.lock().unwrap();
 
         guard.scale_factor
+    }
+
+    /// Access the underlying [`OutputInfo`].
+    ///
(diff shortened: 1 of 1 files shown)


src/output.rs
```diff
@@ -240,11 +240,18 @@ impl OutputData {
...
 
+    /// Get the output scale factor.
     pub fn scale_factor(&self) -> i32 {
         let guard = self.0.lock().unwrap();
 
         guard.scale_factor
     }
+
+    /// Access the underlying [`OutputInfo`].
+    pub fn with_output_info<T, F: FnOnce(&OutputInfo) -> T>(&self, callback: F) -> T {
```
i509VCB, line 251: We should probably warn about panics if you call this reentrantly.

kchibisov: You mean within a callback again?

i509VCB: Yes

kchibisov: And we can't use `mut` for that, since data ruterns non-mut?

kchibisov: I think it won't panic though, just a deadlock will occur?

i509VCB (approved):

PR #337, opened 2023-01-02
pointer: add a way to access the latest_enter_serial

kchibisov:
This should help when using custom cursor images or trying to hide the cursor.

### DIFF
--- a/src/seat/pointer/mod.rs
+++ b/src/seat/pointer/mod.rs
@@ -119,6 +119,13 @@ pub struct PointerData {
     pub(crate) inner: Mutex<PointerDataInner>,
 }
 
+impl PointerData {
+    /// The latest serial from the `Enter` event.
+    pub fn latest_enter_serial(&self) -> Option<u32> {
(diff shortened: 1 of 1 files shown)


src/seat/pointer/mod.rs
```diff
@@ -119,6 +119,13 @@ pub struct PointerData {
     pub(crate) inner: Mutex<PointerDataInner>,
 }
 
+impl PointerData {
+    /// The latest serial from the `Enter` event.
+    pub fn latest_enter_serial(&self) -> Option<u32> {
+       self.inner.lock().unwrap().latest_enter
```
wash2, line 125: it seems like this isn't indented 4 spaces, causing the lint to fail.

wash2 (changes requested):

src/seat/pointer/mod.rs
i509VCB: Yeah cargo fmt picks that up.

wash2 (approved):

PR #338, opened 2023-01-03
keyboard: fix repeat interval

kchibisov:
The interval is in CHARACTERS per second.

### DIFF
--- a/src/seat/keyboard/repeat.rs
+++ b/src/seat/keyboard/repeat.rs
@@ -159,8 +159,8 @@ impl EventSource for KeyRepeatSource {
                                 match info {
                                     // Store the repeat time, using it for the next repeat sequence.
                                     RepeatInfo::Repeat { rate, delay } => {
-                                        // Number of repetitions per second / 1000 ms
-                                        *gap = (rate.get() / 1000) as u64;
+                                        // Use the gap in microseconds.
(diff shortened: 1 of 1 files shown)


wash2:
Ok thanks! I see what you mean. Looking at https://wayland.app/protocols/wayland#wl_keyboard:event:repeat_info it seems like it might be possible for the rate to be 0 though? Do you think it would be good to add a check that disables repetition in the case that `rate` is 0?

kchibisov:
The entire logic wrt keyboard repeat should be cleaned imho. The way it's right now suggests that it was not really tested given that part of the code was using ms, and then doing delays in usec.

I'd look more into it once I have time, the thing is that I simply wanted to open a patch in a way that fixed my problem during winit update.

wash2:
> I'd look more into it once I have time, the thing is that I simply wanted to open a patch in a way that fixed my problem during winit update.

I understand that you only want to fix your problem, but I ask to handle the case of 0 because your current fix might end up dividing by zero, which would panic. Do you think you could make an adjustment to disable repetition in the case of 0?

kchibisov:
>I'd look more into it once I have time

wash2 (approved):
Ok, double checking the type of rate, I can see that it is NonZeroU32 so I'll approve and merge this.

PR #339, opened 2023-01-05
keyboard: use typesafe wrappers for gap and delay

kchibisov:
This will ensure that our time computations are all sane and remove C like random unsigned additions.


wash2 (approved):
Thanks

Issue #292, opened 2022-08-21
Support memfd on FreeBSD platform


HeavyRain266:
FreeBSD also have support for memfd, same as linux. They added it years ago for linux compatibility.

See: https://www.freebsd.org/cgi/man.cgi?query=memfd_create&sektion=3&n=1

i509VCB:
I think this actually needs to be fixed upstream in nix, it only labels android and linux has having memfd

i509VCB:
Actually there is an issue about it: https://github.com/nix-rust/nix/issues/1775

i509VCB:
Just need a new nix release at this point.

PR #341, opened 2023-01-12
Use `memfd_create` on FreeBSD

ids1024:
Fixes https://github.com/Smithay/client-toolkit/issues/292.

### DIFF
--- a/src/shm/raw.rs
+++ b/src/shm/raw.rs
@@ -171,7 +171,7 @@ impl io::Seek for RawPool {
 
 impl RawPool {
     fn create_shm_fd() -> io::Result<RawFd> {
-        #[cfg(target_os = "linux")]
+        #[cfg(any(target_os = "linux", target_os = "freebsd"))]
         {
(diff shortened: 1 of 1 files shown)


ids1024:
Ah, fails to link on FreeBSD 12.1? Well, I guess that answers my question of when this was added. Not sure if there's any easy way to conditionally build based on FreeBSD version (would depending on features from the version of FreeBSD a project is compiled on be expected behavior?).

i509VCB:
Could we try to dynamically load memfd_create on FreeBSD? Or would that be overcomplicated?

ids1024:
Would we need to `dlopen` `libc.so` to do that? A bit awkward.

FreeBSD also has a  `SHM_ANON` flag that can be passed as the path for `shm_open`. Which I think is basically the same as `memfd_create` but without support for sealing.

i509VCB:
One question we could ask here is would a feature make sense to guard this from FreeBSD 12?

PR #342, opened 2023-01-14
feat: simple_layer use esc to exit

Decodetalkers:
Now press esc can exit simple-layer event

### DIFF
--- a/examples/simple_layer.rs
+++ b/examples/simple_layer.rs
@@ -68,6 +68,7 @@ fn main() {
     let layer = LayerSurface::builder()
         .size((256, 256))
         .anchor(Anchor::BOTTOM)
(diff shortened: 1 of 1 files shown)


examples/simple_layer.rs
```diff
@@ -276,6 +278,10 @@ impl KeyboardHandler for SimpleLayer {
         event: KeyEvent,
     ) {
         println!("Key press: {:?}", event);
+        // press 'q' to exit
```
i509VCB, line 281: the comment is incorrect.

Decodetalkers: > the comment is incorrect.

Fixed, thanks

i509VCB (approved):

PR #343, opened 2023-01-20
seat/pointer: fix `ThemedPointer::set_cursor` udata

kchibisov:
The `ThemedPointer` always assumed that it's being used with the `PointerData` leading to issues when used with custom udata.

--

Not sure if that's an ideal solution, maybe you can somehow avoid generic param all-together.


wash2 (approved):
This seems like a good fix to me. Is it ready to merge or did you want to keep working on it?

kchibisov:
I don't have any changes for that one, otherwise I'd open as draft.

PR #344, opened 2023-01-21
Remove `GlobalProxy::NotReady` variant

ids1024:
This doesn't seem to serve any role after https://github.com/Smithay/client-toolkit/pull/313.


wash2 (approved):
lgtm

PR #345, opened 2023-01-21
Add relative pointer, and an example using it

ids1024:
In a future PR adding pointer constraints, the example could be updated to demonstrate relative motion in combination with pointer locks and confinement. But here it just prints motion events a window receives.


wash2 (approved):
Looks good to me. I guess there probably isn't much use for generic user data here?

ids1024:
Probably not. It might make sense for consistency, but I think the data you'd want to associate with the relative pointer is mainly what `wl_pointer` it was created with, which this implementation handles.

PR #347, opened 2023-01-25
Add bindings for pointer constraints 

ids1024:
This needs documentation, and the example can be updated to demonstrate its usage.

Not sure if we can provide any useful higher-level API over `ZwpConfinedPointerV1` and `ZwpLockedPointerV1`. I don't think it's possible to provide an API that can't produce protocol errors at the level sctk operates.


ids1024:
Updated the `relative_pointer` example to show movements and demonstrate pointer locking and confinement.

wash2 (approved):
Looks good, and the example demonstrates use well.

PR #353, opened 2023-02-06
seat: explicitly pass scale each time for pointer

kchibisov:
Previous API flow was to recreate a `ThemedPointer` each time you want to change scale, which is counterintuitive, given that the inner wl_pointer, shouldn't be dropped.


wash2 (approved):

kchibisov:
Rebased.

PR #357, opened 2023-02-26
Refactor: Keyboard Repetition

### LINKED ISSUE #356: Keyboard repeat sometimes blasts
From time to time I've noticed that I get a key repeated multiple times without the key being held for long enough, just arbitrary happens while typing. Maybe the timer invalidation is crewed in some cases?

I don't have clear steps to reproduce other than use alacritty with wayland-rs 0.30 for a day and see things break from time to time...

**This doesn't happen with sctk 0.16.**

wash2:
Refactors keyboard repetition closer to original 0.16 implementation and fixes #356

### DIFF
--- a/src/seat/keyboard/mod.rs
+++ b/src/seat/keyboard/mod.rs
@@ -5,6 +5,7 @@ use std::{
     convert::TryInto,
     env,
     fmt::Debug,
+    marker::PhantomData,
     num::NonZeroU32,
     os::unix::io::AsRawFd,
     sync::{
@@ -14,6 +15,7 @@ use std::{
     time::Duration,
 };
 
+#[cfg(feature = "calloop")]
 use calloop::timer::{TimeoutAction, Timer};
 use wayland_client::{
     protocol::{wl_keyboard, wl_seat, wl_surface},
@@ -321,6 +323,7 @@ pub struct KeyboardData<T> {
     xkb_compose: Mutex<Option<xkb::compose::State>>,
     #[cfg(feature = "calloop")]
     repeat_data: Arc<Mutex<Option<RepeatData<T>>>>,
+    _phantom_data: PhantomData<T>,
 }
 
 impl<T> Debug for KeyboardData<T> {
@@ -366,6 +369,7 @@ impl<T> KeyboardData<T> {
             xkb_compose: Mutex::new(None),
             #[cfg(feature = "calloop")]
             repeat_data: Arc::new(Mutex::new(None)),
+            _phantom_data: PhantomData,
         };
 
         udata.init_compose();
@@ -404,6 +408,7 @@ impl<T> KeyboardData<T> {
             xkb_compose: Mutex::new(None),
             #[cfg(feature = "calloop")]
             repeat_data: Arc::new(Mutex::new(None)),
+            _phantom_data: PhantomData,
         };
 
         udata.init_compose();
--- a/src/seat/keyboard/mod.rs
+++ b/src/seat/keyboard/mod.rs
@@ -9,10 +9,12 @@ use std::{
     os::unix::io::AsRawFd,
     sync::{
         atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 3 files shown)


wash2:
Even if I use an arc mutex to share the repeat key, instead of using a channel to communicate repeat events, as in my last commit, there continue to be spurious repeats. It seems to me that the `Released` event is just not processed until after the timer event goes off, even though it occurred first. I'm not super sure where to go from here, but I'll keep thinking on it :sweat_smile:

kchibisov:
Could we simply remove the timer source and re-arm on press? I think dropping timer source should be faster, no?

For me the events are quite seldom though. It's also strange that `Arc<Mutex<>>` approach isn't working, maybe something is not communicated in time?

wash2:
> Could we simply remove the timer source and re-arm on press? I think dropping timer source should be faster, no?

I'll try that, but I think if I understand what you mean, it should have the same-ish effect that using `Arc<Mutex<>>` does.

wash2:
Ok and I guess a small clarification is, I can't produce a repeat before a release a repeat now even with just 60ms delay after spamming a single key for a while. I can produce a repeat when quickly pressing multiple keys though.

kchibisov:
I guess that's my issue actually, I do press multiple keys near at the same time given that I type fast. I haven't tested any of your recent patches though.

wash2:
I've refactored the key repetition to be more like how it was implemented in 0.16, but with a delay of 60, I can still produce spurious repeats when key-mashing. I don't think the timer is firing at the wrong time, but what seems to happen is that it sometimes blocks until it goes off, preventing other key events from being processed in the meantime?

That being said, I think for now I'll mark this ready for review again :sweat_smile: I think the refactor is still probably an improvement.

wash2:
Ok, removing the ping source seems to fix the spurious repeats. Sorry for spamming so much with this PR.

kchibisov:
>Ok, removing the ping source seems to fix the spurious repeats. Sorry for spamming so much with this PR.

@wash2 I'll add it into my local bind then for testing and report back, thx.

src/seat/keyboard/mod.rs
```diff
@@ -657,13 +670,54 @@ where
...
+                                                    }
+                                                    let key = key.as_mut().unwrap();
+                                                    key.key.time += if key.is_first {
+                                                        key.is_first = false;
+                                                        delay
+                                                    } else {
+                                                        gap.as_millis() as u32
+                                                    };
+                                                    let e = key.key.clone();
+                                                    callback(e, t);
```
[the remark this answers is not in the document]
wash2: Yup!

kchibisov (changes requested):

kchibisov (approved):
Haven't seen any issue while using alacritty for 2 days. So I guess the issue is resolved.

src/seat/keyboard/repeat.rs
```diff
@@ -17,32 +15,28 @@ use super::{
...
-    /// Key has changed during the repeat, but the repeat shouldn't stop.
-    KeyChanged(KeyEvent),
+pub type RepeatCallback<T> = Box<dyn FnMut(&mut T, &WlKeyboard, KeyEvent) + 'static>;
 
-    /// The repeat info has changed.
-    RepeatInfo(RepeatInfo),
+pub(crate) struct RepeatData<T> {
+    pub(crate) current_repeat: Option<RepeatedKey>,
+    pub(crate) repeat_info: RepeatInfo,
+    pub(crate) loop_handle: LoopHandle<'static, T>,
```
i509VCB, line 29: I think this will fail to compile if the `calloop` feature is disabled.

wash2: ya, but the repeat module is feature gated. I did check just now though and it seems another feature gate is missing.

i509VCB (approved):
Seems fine to me at least.

PR #358, opened 2023-02-27
Fix wgpu example

ids1024:
Looks like https://github.com/Smithay/client-toolkit/pull/322 broke this.

I guess the other examples are tested on CI but this one isn't since it requires `--features="wayland-backend/client_system"`?


i509VCB:
Yeah I guess that's an oversight in our current CI build.

ids1024:
It's also a weakness of cargo. It can be unexpected that `cargo build --examples` silently skips examples that have `required-features`.

i509VCB:
Did we want to fix the CI issue or just merge this for now.

vberger (approved):

PR #360, opened 2023-02-28
Don't use raw identifier in method name

ids1024:
Raw identifiers are really only meant for cases where it's technically necessary to use a particular identifier (like when a keyword was added by an edition, but an API currently uses that as an identifier).

I this an `_` suffix is the most idiomatic solution.

### DIFF
--- a/src/shell/xdg/window/mod.rs
+++ b/src/shell/xdg/window/mod.rs
@@ -264,7 +264,7 @@ impl Window {
         }
     }
 
-    pub fn r#move(&self, seat: &wl_seat::WlSeat, serial: u32) {
+    pub fn move_(&self, seat: &wl_seat::WlSeat, serial: u32) {
         self.xdg_toplevel()._move(seat, serial)
     }
 


i509VCB:
I'm a bit uncertain whether the raw identifier is considered to be idiomatic or if an underscore suffix is more or less idiomatic. Right not it seems to be based on opinion. I'd be interested in seeing how many crates use the move_ style.

Now of course there are some cases where there seems to be an implicit agreement on conflicting names (take type => ty for example).

Another thing I'd like to consider is whether move is not descriptive enough and maybe a more explicit name might help?

ids1024:
I'm not sure I've seen other libraries use raw identifiers like this, and it's potentially an unfamiliar syntax, while `move_` seems more clear if still ugly.

> Another thing I'd like to consider is whether move is not descriptive enough and maybe a more explicit name might help?

There's possibly some sense in sticking with the same name the Wayland protocol uses, but `xdg_toplevel::move` isn't a very clear name. One expects a function like that to take `x`/`y` arguments for where to move the toplevel to... but that's not how xdg shell works.

`begin_move` or `start_move` might not be especially clearer. Looking at the summary in the Wayland protocol spec, `start_interactive_user_driven_move` would be clear... but rather verbose.

Same considerations with `resize`.

kchibisov:
If it matters, the scanner generates the code with underscores, so it's sort of more natural...

i509VCB (approved):
Begrudgingly sure I guess.

PR #362, opened 2023-03-02
window: add client side decorations interface

### LINKED ISSUE #273: 0.30 client side decorations for xdg shell

kchibisov:
While the new decorations look the same as the ones before, the
approach in drawing them was changed, and now decorations are more
instant mode, meaning that you create and drop them when needed.

Fixes #273.

--

This series also adds subcompositor and subsurface handlers. The parent surface stuff is
intended to be used from inside winit, yet, I haven't added it into it.

The example which is using CSD decorations is `themed_window`.

All the drawing was ported from before.

cc @PolyMeilex

### DIFF
--- a/src/shell/xdg/frame/fallback_frame.rs
+++ b/src/shell/xdg/frame/fallback_frame.rs
@@ -1,0 +1,696 @@
+//! The default fallback frame which is intended to show some very basic derocations.
+
+use std::sync::Arc;
+use std::{error::Error, num::NonZeroU32};
+
+use crate::reexports::client::{
+    protocol::{wl_shm, wl_subsurface::WlSubsurface, wl_surface::WlSurface},
+    Dispatch, Proxy, QueueHandle,
+};
+use crate::reexports::protocols::xdg::shell::client::xdg_toplevel::ResizeEdge;
+
+use crate::{
+    compositor::SurfaceData,
+    shell::xdg::{WindowManagerCapabilities, WindowState},
+    shell::WaylandSurface,
+    shm::{slot::SlotPool, Shm},
+    subcompositor::{SubcompositorState, SubsurfaceData},
+};
+
+use super::{DecorationsFrame, FrameAction, FrameClick};
+
+/// The size of the header bar.
+const HEADER_SIZE: u32 = 24;
+
+/// The size of the border.
+const BORDER_SIZE: u32 = 4;
+
+const HEADER: usize = 0;
+const TOP_BORDER: usize = 1;
+const RIGHT_BORDER: usize = 2;
+const BOTTOM_BORDER: usize = 3;
+const LEFT_BORDER: usize = 4;
+
+const BTN_ICON_COLOR: u32 = 0xFFCCCCCC;
+const BTN_HOVER_BG: u32 = 0xFF808080;
+
+const PRIMARY_COLOR_ACTIVE: u32 = 0xFF3A3A3A;
+const PRIMARY_COLOR_INACTIVE: u32 = 0xFF242424;
+
+/// The default ugly frame.
+#[derive(Debug)]
+pub struct FallbackFrame<State> {
+    /// The parent surface.
+    parent: WlSurface,
+
+    /// The latest window state.
+    state: WindowState,
+
+    /// The wm capabilities.
+    wm_capabilities: WindowManagerCapabilities,
+
+    /// Whether the frame is resizable.
+    resizable: bool,
+
+    /// Whether the frame is waiting for redraw.
+    dirty: bool,
+
+    /// The location of the mouse.
+    mouse_location: Location,
+
+    /// The location of the mouse.
+    mouse_coords: (i32, i32),
+
+    /// The frame rendering data. When `None` the frame is hidden.
+    render_data: Option<FrameRenderData>,
+
+    /// The frame queue handle.
+    queue_handle: QueueHandle<State>,
+
+    /// The memory pool to use for drawing.
+    pool: SlotPool,
+
+    /// The subcompositor.
+    subcompositor: Arc<SubcompositorState>,
+
+    /// Buttons state.
+    buttons: [Option<UIButton>; 3],
+}
+
+impl<State> FallbackFrame<State>
+where
+    State: Dispatch<WlSurface, SurfaceData> + Dispatch<WlSubsurface, SubsurfaceData> + 'static,
+{
+    pub fn new(
+        parent: &impl WaylandSurface,
+        shm: &Shm,
+        subcompositor: Arc<SubcompositorState>,
+        queue_handle: QueueHandle<State>,
+    ) -> Result<Self, Box<dyn Error>> {
+        let parent = parent.wl_surface().clone();
+        let pool = SlotPool::new(1, shm)?;
+        let render_data = Some(FrameRenderData::new(&parent, &subcompositor, &queue_handle));
+
+        let wm_capabilities = WindowManagerCapabilities::all();
+        Ok(Self {
+            parent,
+            resizable: true,
+            state: WindowState::empty(),
+            wm_capabilities,
+            dirty: true,
+            pool,
+            queue_handle,
+            subcompositor,
+            render_data,
+            mouse_location: Location::None,
+            mouse_coords: (0, 0),
+            buttons: Self::supported_buttons(wm_capabilities),
+        })
+    }
+
+    fn supported_buttons(wm_capabilities: WindowManagerCapabilities) -> [Option<UIButton>; 3] {
+        let maximize = wm_capabilities
+            .contains(WindowManagerCapabilities::MAXIMIZE)
+            .then_some(UIButton::Maximize);
+        let minimize = wm_capabilities
+            .contains(WindowManagerCapabilities::MINIMIZE)
+            .then_some(UIButton::Minimize);
+        [Some(UIButton::Close), maximize, minimize]
+    }
+
+    fn precise_location(
+        buttons: &[Option<UIButton>],
+        old: Location,
+        width: u32,
+        x: f64,
+        y: f64,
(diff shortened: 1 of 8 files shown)


src/shell/xdg/window/mod.rs
```diff
@@ -208,8 +208,8 @@ impl Window {
         decoration.data::<WindowData>().and_then(|data| data.0.upgrade()).map(Window)
     }
 
-    pub fn show_window_menu(&self, seat: &wl_seat::WlSeat, serial: u32, position: (u32, u32)) {
-        self.xdg_toplevel().show_window_menu(seat, serial, position.0 as i32, position.1 as i32);
+    pub fn show_window_menu(&self, seat: &wl_seat::WlSeat, serial: u32, position: (i32, i32)) {
```
kchibisov, line 211: That was wrong, it's i32, because relative to main surface, so could be negative.

kchibisov:
You might notice, that `set_hidden` is missing, and the reason for it is that you want to hide CSD, you can simply drop the entire frame all together.

i509VCB:
> You might notice, that `set_hidden` is missing, and the reason for it is that you want to hide CSD, you can simply drop the entire frame all together.

For more heavyweight decorations it seems wasteful to throw away then start from scratch with internal state (the decorations might use EGL). Maybe a compromise for that scenario would be a way to detach decorations from a window and then reattach the decorations?

kchibisov:
>For more heavyweight decorations it seems wasteful to throw away then start from scratch with internal state (the decorations might use EGL). Maybe a compromise for that scenario would be a way to detach decorations from a window and then reattach the decorations?

I mean why you should even consume resources when you were asked to use `ServerSide` again? Right now the API enforces `shm`, so it's not like you want to use EGL for them, if we want to allow using something other than shm for decorations the `init` should be different, and likely not attached to the frame, then in this case you might want `hide` actually.

src/shell/xdg/frame/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+
+    /// Create new decorations frame.
+    fn new<State>(
+        parent: &impl WaylandSurface,
+        shm: &Shm,
+        subcomositor: &SubcompositorState,
+        queue_handle: &QueueHandle<State>,
+    ) -> Result<Self, Box<dyn Error>>
+    where
+        State: Dispatch<WlSurface, SurfaceData> + Dispatch<WlSubsurface, SubsurfaceData> + 'static;
```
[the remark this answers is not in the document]
kchibisov: I also think that this should be outside the `trait` all together, then `hidden` method would be needed to not recreate it on the fly. The amount to drop could be done internally by the client.

src/shell/xdg/frame/fallback_frame.rs
```diff
@@ -0,0 +1,595 @@
...
+        let buttons = [UIButton::Close, UIButton::Maximize, UIButton::Minimize];
+        for (idx, button) in buttons.into_iter().enumerate() {
+            if width >= (idx + 1) as u32 * HEADER_SIZE {
+                if is_active && mouse_location == &Location::Button(button) {
+                    Self::draw_button(
+                        canvas,
+                        idx * HEADER_SIZE as usize,
+                        scale,
+                        width as usize,
+                        BTN_HOVER_BG.to_ne_bytes(),
```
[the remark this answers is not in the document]
kchibisov: I also had this concern while porting, given that I remember that it should be `le` everywhere.

kchibisov:
I've moved constructor, added back `is_hidden` and `set_hidden`.

kchibisov:
I wonder whether we should maybe put it in a separate crate somehow, the current interface has very few connections with Wayland what so ever. Though, Like the trait by itself doesn't force anything specific, other than asking for `WlSurface` in one place and for `WindowStates` in other, though the window states could be refactored all together.

What do you think? Though, I'm not sure how compatible it could be though.

src/shell/xdg/frame/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::reexports::protocols::xdg::shell::client::xdg_toplevel::{
+    ResizeEdge, State as WindowState,
+};
+
+pub mod fallback_frame;
+
+/// The interface for the client side decorations.
+pub trait DecorationsFrame: Sized {
+    /// Configurations for this frame.
+    type Config;
```
i509VCB, line 13: I'd argue the config doesn't really need to be part of the frame trait. The config is going to be dependent on the frame type anyways.

src/shell/xdg/frame/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn on_click(&mut self, click: FrameClick, pressed: bool) -> Option<FrameAction>;
+
+    /// Emulate pointer moved event on the decorations frame.
+    ///
+    /// The `x` and `y` are location in surface local coordinates relative to `surface`.
+    /// This function returns the new cursor icon you should apply to provide better visual
+    /// effect for the end user.
+    fn click_point_moved(&mut self, surface: &WlSurface, x: f64, y: f64) -> Option<&str>;
+
+    /// All clicks left the decorations.
```
i509VCB, line 31: Clarify what was meant here?

kchibisov: it's simply to move pointer away from the decoration frame.

src/shell/xdg/frame/mod.rs
kchibisov: I think the goal is to transparently pass it, we might remove it, sure?

kchibisov:
I don't think I have anything else left for this.

src/shell/xdg/frame/mod.rs
i509VCB: Isn't moving the pointer off the frame possible even without clicking? I do understand that clicking and the client not doing an interactive resize or move is possible.

I could see more advanced decorations possibly highlighting parts of the frame if the pointer went over some point if not clicking.

src/shell/xdg/frame/mod.rs
```diff
@@ -0,0 +1,133 @@
...
+    ///
+    /// The size must be without the borders, as in [`Self::subtract_borders]` were used on it.
+    ///
+    /// **Note:** The [`Self::update_state`] and [`Self::update_wm_capabilities`] **must be**
+    /// applied before calling this function.
+    ///
+    /// # Panics
+    ///
+    /// Panics when resizing the hidden frame.
+    fn resize(&mut self, width: u32, height: u32);
```
i509VCB, line 60: What should occur if the caller passes 0 for both the width and height

src/subcompositor.rs
```diff
@@ -0,0 +1,124 @@
...
+    D: Dispatch<WlSubsurface, SubsurfaceData>,
+{
+    fn event(
+        _: &mut D,
+        _: &WlSubsurface,
+        _: <WlSubsurface as Proxy>::Event,
+        _: &SubsurfaceData,
+        _: &Connection,
+        _: &QueueHandle<D>,
+    ) {
```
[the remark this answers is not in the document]
kchibisov: Yes, no events.

src/shell/xdg/frame/mod.rs
kchibisov: That's a good question, do we want `NonZero`?

src/shell/xdg/frame/mod.rs
kchibisov: >Isn't moving the pointer off the frame possible even without clicking?

Ah, guess comment is what actually confuses you? because it should probably be `all click points` left the decorations.

Be aware that I use `click` terminology to show that we're not talking about `wl_pointer` events in the first place, you could use all other sorts of inputs with it, like touch.

The use case for such API is to simply `leave` so `Hovered` state will go away, for example when clicking on minimize button and then restoring back.

src/shell/xdg/frame/mod.rs
wash2: I think `NonZero` would be good here.

kchibisov: I'd suggest for configure to pass `NonZeroU32` as well then, because it's actually a `NonZeroU32` already.

wash2: https://wayland.app/protocols/xdg-shell#xdg_toplevel:event:configure

I think a configure event may set width or height to 0 if the client should "decide its own window dimenstions"

kchibisov: I've updated to use `NonZeroU32`.

kchibisov:
ping.

src/shell/xdg/frame/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+
+use crate::reexports::client::protocol::wl_surface::WlSurface;
+use crate::reexports::protocols::xdg::shell::client::xdg_toplevel::ResizeEdge;
+
+use crate::shell::xdg::window::{WindowManagerCapabilities, WindowState};
+
+pub mod fallback_frame;
+
+/// The interface for the client side decorations.
+pub trait DecorationsFrame: Sized {
```
i509VCB, line 13: Decorations feels redundant in the name imo

src/subcompositor.rs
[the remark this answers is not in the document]
i509VCB: In that case an unreachable would be nice in that function.

i509VCB (approved):
This seems ready minus some minor commenfs

src/shell/xdg/frame/mod.rs
kchibisov: I feel like you'd need to rename the trait most of the time, so I've added the prefix, if anything, it's in a `frame`, not in `decorations` module.

wash2 (approved):

PR #366, opened 2023-03-08
window: use bitflags for window state

kchibisov:
This should simplify handling and storing the state downstream.


kchibisov:
I've added `v5` of the `xdg_shell` as well in a separate commit.

**This branch must be rebased instead of squashed**

wash2 (approved):

PR #369, opened 2023-03-24
Use map_copy_read_only() to map the keymap file

yshui:
map() creates a shared, read-only map. Unfortunately there's a quirk in Linux that prevents the creation of such maps if the file is sealed for write. (which is a reasonable thing the compositor might want to do).

map_copy_read_only() creates a private, read-only map which doesn't have this problem.

### DIFF
--- a/src/seat/keyboard/state.rs
+++ b/src/seat/keyboard/state.rs
@@ -361,7 +361,7 @@ impl KbState {
     }
 
     pub(crate) unsafe fn init_with_fd(&mut self, fd: File, size: usize) {
-        let map = MmapOptions::new().len(size).map(&fd).unwrap();
+        let map = MmapOptions::new().len(size).map_copy_read_only(&fd).unwrap();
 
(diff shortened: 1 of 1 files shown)


wash2:
Thanks, I think a similar change was made in the xkbcommon crate a while back too. https://github.com/rust-x-bindings/xkbcommon-rs/commit/bd3aeaa84b03cebc3eed8f65b591c7adc9cf5411 

Do you need this fix for something you are working on with sctk 0.16? sctk 0.30 should be released relatively soon.

yshui:
Not with sctk, I was working on a compositor when I noticed if I add write seal to the keymap file, sctk would panic.

I am happy to close this if sctk 0.30 is going to cover it.

kchibisov:
>I am happy to close this if sctk 0.30 is going to cover it.

You can try running example from the latest master, like `cargo run --example simple_window`.

yshui:
yep, it works

PR #372, opened 2023-03-28
Data offer source actions

wash2:
While the description at https://wayland.app/protocols/wayland#wl_data_offer:event:source_actions is confusing to me, the way I read it implies that enter event should be sent before source actions for DnD offers. In practice, I've only ever seen the source actions sent before the enter event :sweat:.

This PR fixes the implementation so that an undetermined offer also tracks advertised source actions. When the enter event is being handled, source actions can be checked and handled as well at the same time. It doesn't call the source_actions method though, because the method expects a drag offer. It also cleans up an unused variable and makes a couple methods take `&self` instead of `&mut self`.

Ultimately I don't think this event matters so much because clients can still set their preferred action and accepted actions regardless, but the info from the source actions should be available if it is received.

### DIFF
--- a/examples/data_device.rs
+++ b/examples/data_device.rs
@@ -500,7 +500,7 @@ impl DataDeviceWindow {
 
 impl DataDeviceHandler for DataDeviceWindow {
     fn enter(&mut self, _conn: &Connection, _qh: &QueueHandle<Self>, data_device: DataDevice) {
-        let mut drag_offer = data_device.drag_offer().unwrap();
+        let drag_offer = data_device.drag_offer().unwrap();
         println!("data offer entered x: {:.2} y: {:.2}", drag_offer.x, drag_offer.y);
 
         // accept the first mime type we support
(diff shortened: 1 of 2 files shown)


i509VCB:
> enter event should be sent before source actions for DnD offers

I am curious what other clients expected behavior is regarding that. I'd at least like to know if this is a case of the protocol being unspecified and if that is the case then report it upstream.

wash2:
It looks like gtk does something similar to what I'm doing here. If they haven't received the enter event they track the source actions on a pending offer in the seat, but otherwise if there is a known drop, they make the change to the drop and emit a motion event. 

The change was added here.
https://gitlab.gnome.org/GNOME/gtk/-/commit/7f8a8f221d60bc5da23602e5e024eb91f8b0f758

It still looks pretty much the same.
https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gdk/wayland/gdkseat-wayland.c#L258

i509VCB (approved):

PR #375, opened 2023-04-05
fix: delegate dispatch for ZxdgDecorationManagerV1 in delegate_xdg_shell

### LINKED ISSUE #374: delegate_xdg_shell does not delegate Dispatch for ZxdgDecorationManagerV1
Currently this is done in `delegate_xdg_window`. However a client which only creates for popups for the layer shell as an example will need to use `delegate_xdg_window`.

wash2:
fixes #374


elinorbgr (approved):

i509VCB:
get_toplevel_decoration is pretty clear that only toplevels can have xdg-decoration

i509VCB (approved):

PR #378, opened 2023-04-24
Use xkeysym for keyboard handling

notgull:
Replaces the `keysyms` module with `xkeysym` and makes APIs take `xkeysym::Keysym` instead of alternatives. See notgull/xkeysym#6


kchibisov:
Could you rebase with the current latest version?

Cargo.toml
```diff
@@ -31,8 +31,9 @@ wayland-protocols-wlr = { version = "0.1.0", features = ["client"] }
 wayland-scanner = "0.30.0"
 wayland-csd-frame = { version = "0.2.2", default-features = false, features = ["wayland-backend_0_1"] }
 
-xkbcommon = { version = "0.5", optional = true, features = ["wayland"] }
+xkbcommon = { version = "0.5.0", optional = true, features = ["wayland"], git = "https://github.com/notgull/xkbcommon-rs.git", branch = "xkeysym" }
 calloop = { version = "0.10.5", optional = true }
+xkeysym = { git = "https://github.com/notgull/xkeysym.git", version = "0.1.1" }
```
kchibisov, line 36: It's a bit weird that you've picked it like that, we have a stable release with the good version. And the crate is published.

kchibisov:
And I guess it's not a draft anymore at this point?

notgull:
> And I guess it's not a draft anymore at this point?

Unfortunately I can't take it out of draft status until https://github.com/rust-x-bindings/xkbcommon-rs/pull/43 is released.

kchibisov:
@notgull I mean, we have a release we can use, I don't see any issue with using it here given that it's better than hand rolled solution and it's already used in winit, so we have less code size upstream.

notgull:
Oh, are you talking about `xkbcommon-dl`? Sounds good I'll port to that

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 #### Breaking Changes
 
 - `wayland-rs` dependencies are updated to 0.30 and all APIs have changed significantly as a result
+- `xkeysym::Keysym` is used as a keyboard key representation instead of `u32`
```
kchibisov, line 32: You should put it in the right place.

notgull: This is a breaking change, I think?

kchibisov: I haven't said that it's not, it just for 0.17 which was released months ago. You should move to the actual unreleased section.

notgull: My bad, fixed!

PolyMeilex (approved):
Looks great to me.

EDIT:
Looks like it's not possible to use those with match statements?

wash2 (approved):

PolyMeilex:
Thanks!

PR #381, opened 2023-05-06
seat/pointer: automatically use cursor-shape-v1

kchibisov:
When using `ThemedPointer` automatically use the `cursor-shape-v1`
protocol to set cursor icon.

We still accept the `WlSurface` and `WlShm` when building the
`ThemedPointer` to handle the pointer hiding and custom cursor
setting.

--

The protocol could be found https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/194 .

I'm not sure if abusing `SeatState` is the way to go, but I don't really like passing the global around(messy) or implicitly binding(some clients don't use pointer). So for now I've decided to optionally pass the value to `::new`.

The macros handling is baked into `delegate_pointer` because the protocol doesn't have events on its own and there's nothing to customize.

The patch is built on top of #380.


kchibisov:
Blocked on https://github.com/Smithay/wayland-rs/pull/624

src/seat/mod.rs
```diff
@@ -91,11 +94,26 @@ impl SeatState {
...
 
+    /// Provide the [`CursorShapeManager`] to automatically handle server side cursor
+    /// when using [`Self::get_pointer_with_theme`] and [`Self::get_pointer_with_theme_and_data`].
+    pub fn with_cursor_shape_manager(
+        mut self,
+        cursor_shape_manager: Option<CursorShapeManager>,
+    ) -> Self {
+        self.cursor_shape_manager = cursor_shape_manager;
+        self
+    }
```
kchibisov, line 105: Maybe we should lazily bind the manager internally when trying to get a pointer, so users will natively have a cursor shape working?

kchibisov:
beep @wash2 @i509VCB

src/seat/mod.rs
wash2: Ya, I think that would work well in get_pointer_with theme.

kchibisov: Should be fixed now.

wash2 (approved):

PR #382, opened 2023-05-06
chore: use `wayland-csd-frame` for CSD interop

kchibisov:
smithay-client-toolkit could have various breaking changes, to avoid
dealing with the ecosystem fallout the client side decorations
frame was extracted to another repository and provides more or less
stable interface, so client-toolkit could be bumped without
in other libraries using that crate without introducing a breaking
update.

--

The patch is based on #380. 

The crate is located https://github.com/rust-windowing/wayland-csd-frame and not released yet. Releases of both crates are expected to be performed before merging this.


kchibisov:
This should be all good now.

i509VCB (approved):

PR #383, opened 2023-05-16
Add `wp_presentation` protocol

ids1024:
A fairly simple protocol (that basically amounts to a single function returning a future, functionally?). But a couple small things are a bit awkward here.

Also needs example and basic docs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,6 +23,7 @@ pub mod data_device_manager;
 pub mod error;
 pub mod globals;
 pub mod output;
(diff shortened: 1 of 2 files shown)


ids1024:
Well, I could add an example, but it just amounts to `simple_window` but printing a message every frame. Not sure we need that.

PolyMeilex (approved):
Docs would be nice, but I'm just going to merge this as is, for the sake of not having a perfectly fine usable code in a stale MR.

PR #387, opened 2023-06-08
FallbackFrame: make subsurfaces desync

nicolasavru:


nicolasavru:
Sorry, I forgot about that example.

Looking at it, its logic should be fine as-is: every time it receives a frame event, it draws the window contents and decorations, so it's already practically redrawing the decorations as soon as it should. If, on the other hand, the draw loop would block on something else (for something to change in its internal state, for more user input to come in, etc.), then you'd want to call frame.draw() outside of the frame callback.

I'm not sure how to make that clear in the example without significantly complicating it and changing its event loop. Really that should be in the documentation for FallbackFrame, but there isn't much of that yet.

kchibisov:
What's the reason to use `desync`? I find it wrong that those are desync, given that it could result in visual glitches for the user, for example, when doing interactive resizes and your CSD resize before the actual window.

The sync here is intentional and was always used even in previous versions of the sctk(sync is the default).

nicolasavru:
I touch upon this on the code comment, I'll put more detail here. A frame callback means that a surface _may_ draw a new frame, not that it _must_ draw one. A valid strategy for an application is to only draw frames when something has actually changed: it's humming along in its internal event loop, something changes so it decides to draw a new frame, it sees that it has received a frame callback since the last time it drew, so it draws. Then it receives another frame callback a short time later (when the compositor is ready for the next frame), but nothing has changed in its internal state, so it doesn't draw the next frame yet and waits until it needs to. In the meantime, the user's pointer is moving over the minimize/maximize/close buttons in the frame and the frame is drawing its new frames (there's a highlight effect when you mouse over the buttons), but none of those frames will get drawn until the parent surface draws, which may happen at some indeterminate time in the future. So the the mouseover animations just don't work. As as is, the FallbackFrame is only really compatible with applications which re-draw the parent surface on every frame callback.

A strategy I've seen some applications use is to normally have subsurfaces by desync, make sync during resizes, and then make them desync again when the resize is done. I could take a crack at that here if you think that's a good idea.

kchibisov:
>A strategy I've seen some applications use is to normally have subsurfaces by desync, make sync during resizes, and then make them desync again when the resize is done. I could take a crack at that here if you think that's a good idea.

That sounds better, yes. I wonder whether we'll need an explicit API for that or do things internally only, I think it could be done transparently for the user from what I can tell.

nicolasavru:
I'll give that a shot this week. I also think we can do it transparently, we'll see if I hit any blockers there.

nicolasavru:
Apologies, I got lost in the morass of copy/paste and drag-and-drop. Just pushed a new version which has the subsurfaces as normally desync, but sync during resizes.

wash2 (approved):

kchibisov:
I don't think this is correct though. `Resizing` is just a part, you can resize a window without a resizing state.

I don't think this can work without a recent changes in `wayland-csd-frame`, but I'll pull and update things after the `wayland-backend` will be propagated.

nicolasavru:
What other mode of resizing are you thinking of? Just changing the size of the buffer you're committing without being asked to by the compositor? Yeah, not sure we can currently handle that transparently for the user and we'd need an api for them switch the mode when they plan to resize the parent service.

kchibisov:
@nicolasavru it can resize without a mode change, for example when the compositor in tiling.

>Yeah, not sure we can currently handle that transparently for the user and we'd need an api for them switch the mode when they plan to resize the parent service.

I've already added API for all of that, I'm just waiting for a new wayland-client release, so I can pull my crate given that everything will use `wayland-backend 0.2.0`.

nicolasavru:
Got it, we can revisit this after that happens.

kchibisov:
I think this is resolved though? I've made subsurfaces desync in the PR that auto closed this PR.

PR #397, opened 2023-07-31
Add a wp_viewporter implementation

linkmauve:
This simplifies client code quite a bit.

### DIFF
--- a/src/viewporter.rs
+++ b/src/viewporter.rs
@@ -1,0 +1,131 @@
+use crate::reexports::client::globals::{BindError, GlobalList};
+use crate::reexports::client::protocol::wl_surface::WlSurface;
+use crate::reexports::client::{Connection, Dispatch, Proxy, QueueHandle};
+use crate::reexports::protocols::wp::viewporter::client::{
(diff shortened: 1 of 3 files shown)


src/viewporter.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// Set the source rectangle.
+    pub fn set_source(&self, x: f64, y: f64, width: f64, height: f64) {
+        self.viewport.set_source(x, y, width, height);
+    }
+
+    /// Set the destination size.
+    pub fn set_destination(&self, width: i32, height: i32) {
+        self.viewport.set_destination(width, height);
+    }
+    */
```
PolyMeilex, line 100: Can we uncomment those?
It would be good to use this wrapper instead of WlVieporter directly, as that would allow for addition of version guards when a new version of viewporter releases some day.

chrisduerr: Is there any point to `Viewport` to begin with? Just getting the viewport from the state and using that is all you need really. My personal implementations never used more than that (https://github.com/catacombing/kumo/blob/master/src/wayland/protocols/viewporter.rs).

PR #398, opened 2023-08-10
Add support for wl_compositor v6

kchibisov:
Wire up new events instead of using watcher on the output and also handle transform changes.

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -1,3 +1,5 @@
+use std::mem;
+use std::sync::MutexGuard;
 use std::sync::{
     atomic::{AtomicI32, Ordering},
     Arc, Mutex,
@@ -9,7 +11,7 @@ use wayland_client::{
         wl_callback, wl_compositor, wl_output, wl_region,
         wl_surface::{self, WlSurface},
     },
-    Connection, Dispatch, Proxy, QueueHandle,
+    Connection, Dispatch, Proxy, QueueHandle, WEnum,
 };
 
 use crate::{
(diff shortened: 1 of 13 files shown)


src/compositor.rs
```diff
@@ -131,6 +133,11 @@ impl SurfaceData {
         self.scale_factor.load(Ordering::Relaxed)
     }
 
+    /// The scale factor of the output with the highest scale factor.
```
wash2, line 136: I think this comment should say transform instead of scale factor?

src/compositor.rs
```diff
@@ -275,51 +324,54 @@ where
...
-            None => return,
-            Some(factor) if factor == current => return,
-            Some(factor) => factor,
-        };
+        dispatch_surface_state_updates(state, conn, qh, surface, data, inner);
+    }
+}
 
-        data.scale_factor.store(factor, Ordering::Relaxed);
+fn dispatch_surface_state_updates<D, U>(
```
wash2, line 337: Because this method is only called if the version is less than 6, will the surface transform ever be anything other than `wl_output::Transform::Normal`? It seems to default to `wl_output::Transform::Normal` and this method will only set it to `wl_output::Transform::Normal` as far as I can tell from the reduce step, so I'm not sure it needs to do anything with transform. Or have I missed something?

kchibisov: >Because this method is only called if the version is less than 6, will the surface transform ever be anything other than wl_output::Transform::Normal? It seems to default to wl_output::Transform::Normal and this method will only set it to wl_output::Transform::Normal as far as I can tell from the reduce step, so I'm not sure it needs to do anything with transform. Or have I missed something?

It might be really confusing, but reduce doesn't apply when you have 1 element. So for more than 2 elements in array, it'll map to `Normal`, but for exactly one output it'll be the corrent transform.

src/compositor.rs
kchibisov: aye.

wash2 (approved):

PR #401, opened 2023-08-24
Integrate calloop-wayland-source into the sctk repo

DJMcNab:
As discussed on Element at
https://matrix.to/#/!qaZCXtKkCIlfHYRUep:safaradeg.net/$N6AcaX4XpJNXr3xaco5jUPltAXyohG2cndwEHeQCBSM?via=libera.chat&via=matrix.org&via=mozilla.org

I've made a few choices which might need thought:
1) I've added some badges to the calloop-wayland-source readme
2) I've not used workspace.dependencies, because that would increase lower bounds
3) I've not copied over the exact details of the rustfmt.toml and
   ci files (in particular the environment variables)
4) I've updated the MSRV to 1.65 to match calloop-wayland-source

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust: ['1.64.0', 'stable', 'beta']
+        rust: ['1.65.0', 'stable', 'beta']
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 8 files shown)


kchibisov:
nack on this, hard to manage and won't be really welcome in winit, like winit doesn't use half of the SCTK, developing calloop integration side by side with sctk is just slowing things down, because it could evolve faster than sctk is and is intended to be used even with raw waylland-client stuff, which could get ahead of client-toolkit.

If you just can't manage your personal workspace it's not a reason to merge unrelated crates.

Issue #402, opened 2023-09-05
`calloop` dependency out of date


OvermindDL1:
`calloop` is up to 0.11 now, which includes an updated nix dependency as well, any chance this could be updated in a release?

wash2:
Sure, I'll see if I can update it for the next release.

kchibisov:
calloop will be updated, once the github.com/smithay/calloop-wayland-source will have an update, which waits for a new calloop breaking version, which should be soon due to https://github.com/Smithay/calloop/pull/144 .

PR #403, opened 2023-09-08
WIP Update wayland-rs and calloop

ids1024:
`simple_Window` example here works.

Depends on https://github.com/Smithay/wayland-rs/pull/656 and https://github.com/rust-windowing/wayland-csd-frame/pull/10.

Incorporates the calloop source that was part of wayland-client, updated for current wayland-rs and wayland-client. Unless we want that somewhere else?

Updating these crates is a breaking change, so we might as well update APIs to use io safe types. Perhaps rustix could be used in some places.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,20 +23,21 @@ log = "0.4"
 memmap2 = "0.7.0"
 nix = { version = "0.26.1", default-features = false, features = ["fs", "mman"] }
 thiserror = "1.0.30"
-wayland-backend = "0.1.0"
-wayland-client = "0.30.2"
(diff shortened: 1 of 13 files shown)


kchibisov:
>Incorporates the calloop source that was part of wayland-client, updated for current wayland-rs and wayland-client. Unless we want that somewhere else?

You're likely not aware about https://github.com/Smithay/calloop-wayland-source which is also blocked on https://github.com/Smithay/client-toolkit/issues/402

ids1024:
Superseded by https://github.com/Smithay/client-toolkit/pull/404, which is now merged. Closing.

PR #404, opened 2023-09-16
Update calloop and wayland-rs to the latest versions

### LINKED ISSUE #402: `calloop` dependency out of date

kchibisov:
Suggested-by: Ian Douglas Scott <idscott@system76.com>
Fixes #402.

Based on #403, cc @ids1024 @DjMcNab

The calloop wayland source doesn't work and not released yet. https://github.com/Smithay/calloop-wayland-source/pull/1

### DIFF
--- a/examples/data_device.rs
+++ b/examples/data_device.rs
@@ -1,11 +1,13 @@
 use std::{
     convert::TryInto,
-    fs::File,
+    fs::{self, File},
     io::{BufRead, BufReader, Write},
+    os::unix::io::OwnedFd,
     time::Duration,
 };
 
-use calloop::{EventLoop, LoopHandle, RegistrationToken};
+use smithay_client_toolkit::reexports::calloop::{EventLoop, LoopHandle, RegistrationToken};
+use smithay_client_toolkit::reexports::calloop_wayland_source::WaylandSource;
 use smithay_client_toolkit::{
     compositor::{CompositorHandler, CompositorState},
     data_device_manager::{
@@ -43,7 +45,6 @@ use smithay_client_toolkit::{
         Shm, ShmHandler,
     },
 };
-use wayland_backend::io_lifetimes::OwnedFd;
 use wayland_client::{
     globals::registry_queue_init,
     protocol::{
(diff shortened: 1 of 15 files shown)


kchibisov:
Should be good now.

wash2 (approved):

examples/data_device.rs
```diff
@@ -652,6 +653,8 @@ impl DataDeviceHandler for DataDeviceWindow {
                     (o, d, Some(t)) => (o, d, t),
                     _ => return,
                 };
+                // SAFETY: it's safe as long as we don't close the underlying file.
+                let f: &mut fs::File = unsafe { f.get_mut() };
```
ids1024, line 657: This is annoying, though I'm not sure of a better way to do it here or in calloop while preserving IO safety. (And also not over complicating things; like manual syscalls with rustix instead of using `File` methods).

ids1024 (approved):
We could make further (breaking API) changes later to use IO safety in public APIs, probably using Rustix, and not need some of the `unsafe` blocks here. But everything should be good here as far as just updating the dependencies.

examples/data_device.rs
kchibisov: Yes, I've discussed this in #calloop channel, basically the issue is that `NoIoDrop` thing can't be re-wrapped by us...

kchibisov: Oh, with the alternative of doing a `dup`, but there's no way I'll do a `dup` for that.

ids1024: `dup` works, but yeah, not great. Technically also perfectly safe to define a struct implementing `Read` using `rustix::io::read` on the fd... but that would be rather verbose if nothing else.

elinorbgr: Actually, I'm thinking in the case of SCTK, handling pipes from data device would be a place where using the futures/IO adapters from calloop could make quite a bit of sense :thinking:

kchibisov: I'm not sure about Futures, but yeah, if we write a specialized object which simply read/writes maybe it'll be nicer in the end, I just ported the API.

wash2:
Should we add a note to the CHANGELOG as part of this PR? I can increment the version and do a release after merging.

kchibisov:
Hm, sure, I can add.

kchibisov:
I've added that we basically updated them, since it's a reexport and is a breaking change.

wash2 (approved):

PR #406, opened 2023-09-16
Restrict SimpleGlobal bindings to our max API version

danieldg:
This is a partial fix for the way SimpleGlobal binds to the latest version offered by the server instead of the latest version the client understands, which causes errors (usually leading to crashes) when a server actually uses the later version and sends an event we don't know about.

This fixes crashes in a number of clients using SCTK-0.16 on compositors supporting the newest wayland surface APIs (to be precise, any that send a `preferred_buffer_scale` event).

### DIFF
--- a/src/environment.rs
+++ b/src/environment.rs
@@ -283,6 +283,7 @@ impl<I: Interface + Clone + From<Proxy<I>> + AsRef<Proxy<I>>> GlobalHandler<I> f
         version: u32,
         _: DispatchData,
     ) {
+        let version = I::VERSION.min(version);
         self.global = Some((*registry.bind::<I>(version, id)).clone())
     }
     fn get(&self) -> Option<Attached<I>> {


kchibisov:
@wash2 could we make it into patch release? It affects wezterm from what I understood since they haven't updated yet and I'm not sure whether they will.

kchibisov:
The CI failures are not relevant if it matters.

src/environment.rs
```diff
@@ -283,6 +283,7 @@ impl<I: Interface + Clone + From<Proxy<I>> + AsRef<Proxy<I>>> GlobalHandler<I> f
         version: u32,
         _: DispatchData,
     ) {
+        let version = I::VERSION.min(version);
```
i509VCB, line 286: If anything I'd want a sort of compile time assertion or an unconditional panic if you try to bind a version above the API one.

danieldg: You can't do that from here.  The version passed in to this function is the maximum version the server supports; there is no information about the version the client knows about other than the maximum version our XML knows about.

kchibisov: Yeah, be aware that it's a 0.16.0, it's a bit different.

i509VCB: Yeah I noticed that after the review comment, I'm less sure about 0.16

kchibisov: How it's done is the only way to do as far as I remember. The alternative is to explicitly bind the version we support though, not just the one maximum in XML we use.

@danieldg maybe we should just bind the actual version we support and optionally clam with the interface one? It should be more robust in case `I::VERSION` will change.

danieldg: What is "the version we support", and how do you get access to it here in the SimpleGlobal impl?

Note that `I::VERSION` shouldn't change unless the wayland-client repository updates its internal wayland.xml dependency.

kchibisov: I just don't remember whether the `wayland.xml` update was a breaking change or not. I think it's not a breaking change as long as enum versions are the same.

You're correct, we can't do anything with `SimpleGlobal` here due to how it's all designed and not do a breaking change.

kchibisov (approved):

src/environment.rs
danieldg: Basically this patch is not a good way to do it if wayland.xml does change (although it would actually fix the error message that clients are hitting, they'll likely hit the  `_ => unreachable!()` branch of an event-handling enum written when the enums had fewer variants.  Thanks to clippy, ignoring unknown variants requires annotations but panicking on them does not.

kchibisov: Yeah, but I don't think the wayland.xml of wayland 0.29 will be updated. We just simply don't have other option as of now.

wash2:
Should we increment the version to 0.16.1 in this PR?

kchibisov:
@wash2 in the separate.

wash2 (approved):

cole-h:
Just a minor note -- it seems that the 0.16.1 tag was never pushed to this repo (though that version has been published on crates.io and docs.rs successfully).

kchibisov:
@cole-h good catch.

@wash2 something like that should do it

```
git tag v0.16.1 -m "smithay-client-toolkit version 0.16.1"
git push origin v0.16.x tag v0.16.1
```

wash2:
Ok the tag is pushed now

cole-h:
Cheers, thanks!

PR #408, opened 2023-09-22
Take `OwnedFd` instead of `RawFd` as argument to `receive_to_fd`

ids1024:
This appears to be the only use of `RawFd` in a public API. With this change, these functions should be safe to call.

I'm not sure if it would make more sense for these functions to not close the file descriptor. In which case they could take `BorrowedFd`. But this matches the current implementation, but with an IO safe type.


wash2 (approved):

ids1024:
Ah, I guess it does make sense to use `OwnedFd` rather than `BorrowedFd` since it's expected to be used with pipes, waiting for EOF on the receiver. Which would depend on not having another copy of the sender held open. So this API should be appropriate.

PR #409, opened 2023-09-22
fix(wl_data_device): drop and leave event handling

wash2:
While the spec says to destroy the wl_data_offer after wl_data_device::leave, I think compositors often send the leave event right away after a drop. This is a bit problematic if an application hasn't finished receiving the data and then the data offer is destroyed before it can make the finish request. I believe this can currently be recreated in the data_device example. The source never receives a finish event because the offer was destroyed and isn't alive by the time we make the request.

This PR tries to avoid destroying an offer after the leave event if a drop event already occurred. It also tries to avoid making request besides finish or destroy.

### DIFF
--- a/examples/data_device.rs
+++ b/examples/data_device.rs
@@ -587,7 +587,7 @@ impl DataDeviceHandler for DataDeviceWindow {
     }
 
     fn leave(&mut self, _conn: &Connection, _qh: &QueueHandle<Self>, _data_device: &WlDataDevice) {
-        println!("data offer left");
+        println!("Data offer left");
     }
 
     fn motion(
@@ -739,6 +739,7 @@ impl DataDeviceHandler for DataDeviceWindow {
                             loop_handle.remove(token);
                             println!("Dropped data: {:?}", String::from_utf8(data.clone()));
                             offer.finish();
+                            offer.destroy();
                             state.dnd_offers.push((offer, Vec::new(), None));
                             return;
                         } else {
@@ -754,6 +755,7 @@ impl DataDeviceHandler for DataDeviceWindow {
                     Err(e) => {
                         eprintln!("Error reading dropped data: {}", e);
                         offer.finish();
+                        offer.destroy();
                         loop_handle.remove(token);
 
                         return;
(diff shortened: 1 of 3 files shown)


kchibisov (changes requested):
We should probably document that the users should destroy the offer? It would also be nice to get which compositors do what you say in particular, but just for the reference.

examples/data_device.rs
```diff
@@ -860,6 +862,7 @@ impl DataSourceHandler for DataDeviceWindow {
         _qh: &QueueHandle<Self>,
         source: &wayland_client::protocol::wl_data_source::WlDataSource,
     ) {
+        println!("FINISHED");
```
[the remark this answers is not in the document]
wash2: O just copied the message above it, but I can change how both messages are written.

wash2:
> It would also be nice to get which compositors do what you say in particular, but just for the reference.

So far I was able to recreate the issue in Cosmic, sway, and Pop on wayland (Gnome), but I can double check vanilla Gnome too

kchibisov:
>So far I was able to recreate the issue in Cosmic, sway, and Pop on wayland (Gnome), but I can double check vanilla Gnome too

That's more than enough I guess, just mention that in the commit. We could probably use some RAII so users won't worry about that, but I'm not familiar with dnd myself much and can just look into the code.

src/data_device_manager/data_offer.rs
```diff
@@ -295,6 +311,31 @@ impl DataOfferData {
         }
     }
 
+    pub(crate) fn drop_performed(&self) {
+        let mut inner = self.inner.lock().unwrap();
+
+        if let DataDeviceOffer::Drag(ref mut o) = inner.deref_mut().offer {
+            o.dropped = true;
+        }
+    }
```
kchibisov, line 320: Do we really need this as a separate function if it's used only once in `Drop` handler?

src/data_device_manager/data_offer.rs
```diff
@@ -106,14 +112,20 @@ impl DragOffer {
     /// This request may happen multiple times for different mime types, both before and after wl_data_device.drop.
     /// Drag-and-drop destination clients may preemptively fetch data or examine it more closely to determine acceptance.
     pub fn receive(&self, mime_type: String) -> std::io::Result<ReadPipe> {
-        receive(&self.data_offer, mime_type)
+        if !self.left || self.dropped {
+            receive(&self.data_offer, mime_type)
```
kchibisov, line 116: Could you document why it's like that?

wash2: I've added a comment that explains why

kchibisov (approved):

elinorbgr (approved):

PR #410, opened 2023-09-22
Use `rustix` instead of `nix`

ids1024:
This is cleaner than `nix` due to the use of `OwnedFd`/`BorrowedFd`.

`calloop` is already using `rustix`.

Depends on https://github.com/bytecodealliance/rustix/pull/848.


ids1024:
0.38.15 is now released with the required functionality.

wash2 (approved):

PR #416, opened 2023-10-08
data_device: return `PostAction` for the Read/Write pipe

kchibisov:
This should simplify client handling, since they can properly model the read/write loop.


kchibisov:
2 commits are independent.

Don't squash on merge.

wash2 (approved):

PR #419, opened 2023-10-12
Add session-lock protocol

ids1024:
I modelled the API here on how layer shell is handled.

`SessionLock` and `SessionLockSurface` destroy their corresponding Wayland objects on drop. In the case of `SessionLock`, it is incorrect to do this before `locked`/`finished` are received, but also incorrect to not destroy it. For security, the compositor remains locked if any of these things are done incorrectly.

Ideally `SessionLock` then might be a linear type that you can't drop but have to pass to an explicit destroy function, but in lieu of such functionality, it's marked `#[must_use]`.

The example simply creates a lock, creates a lock surface for each output with the same colorful graphic as `simple_window`, and terminates after 5 seconds.

### DIFF
--- a/src/session_lock/mod.rs
+++ b/src/session_lock/mod.rs
@@ -1,0 +1,211 @@
+use crate::{compositor::Surface, error::GlobalError, globals::GlobalData, registry::GlobalProxy};
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc, Weak,
+};
+use wayland_client::{
+    globals::GlobalList,
+    protocol::{wl_output, wl_surface},
+    Connection, Dispatch, Proxy, QueueHandle,
+};
+use wayland_protocols::ext::session_lock::v1::client::{
+    ext_session_lock_manager_v1, ext_session_lock_surface_v1, ext_session_lock_v1,
+};
+
+mod dispatch;
+
+/// Handler trait for session lock protocol.
+pub trait SessionLockHandler: Sized {
+    /// The session lock is active, and the client may create lock surfaces.
+    fn locked(&mut self, conn: &Connection, qh: &QueueHandle<Self>, session_lock: SessionLock);
+
+    /// Session lock is not active and should be destroyed.
+    ///
+    /// This may be sent immediately if the compositor denys the requires to create a lock,
+    /// or may be sent some time after `lock`.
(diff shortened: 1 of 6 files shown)


YaLTeR:
Not sure if it's an error in this PR or in my current WIP compositor session-lock impl, but sometimes the unlock request is not received. On the screenshot, the top-left terminal is the compositor's, and the right terminal is the session-lock example's.

image

YaLTeR:
Looks like swaylock does a wl_display.sync() in the end to be sure.

ids1024:
Hm, I thought doing a `flush` would be sufficient. I guess if the compositor reaches EOF reading from the socket, it drops the client and doesn't disconnected queued clients?

`sync` is specifically mentioned in the protocol spec, I see.

> Note that a correct client that wishes to exit directly after unlocking the session must use the wl_display.sync request to ensure the server receives and processes the unlock_and_destroy request. Otherwise there is no guarantee that the server has unlocked the session due to the asynchronous nature of the Wayland protocol. For example, the server might terminate the client with a protocol error before it processes the unlock_and_destroy request.

src/session_lock/mod.rs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    pub fn lock<D>(&self, qh: &QueueHandle<D>) -> Result<SessionLock, GlobalError>
+    where
+        D: Dispatch<ext_session_lock_v1::ExtSessionLockV1, SessionLockData> + 'static,
+    {
+        let session_lock_manager = self.session_lock_manager.get()?;
+
+        // Freeze the queue during the creation of the Arc to avoid a race between events on the
+        // new objects being processed and the Weak in the PopupData becoming usable.
```
wash2, line 177: I guess this comment is supposed to say `SessionLockData`?

ids1024: Ah right.

Though I copied this from `src/shell/wlr_layer/mod.rs`, so it was already a copy-paste error there. And in `src/shell/xdg/mod.rs` as well, it looks like.

I guess there's not a clear way to abstract this logic rather than copying it...

ids1024:
Running several times in Sway, using a `round_trip` instead of a `flush` seems to work reliably.

src/session_lock/mod.rs
ids1024: Updated the comment here, and in the other places.

wash2 (approved):

PR #420, opened 2023-10-18
Macro build fixes

danieldg:
This is suitable for inclusion in a 0.18.1 release.

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -42,8 +42,14 @@ pub trait CompositorHandler: Sized {
 
     /// A frame callback has been completed.
     ///
-    /// This function will be called after sending a [`WlSurface::frame`](wl_surface::WlSurface::frame) request
-    /// and committing the surface.
+    /// Frame callbacks are used to avoid updating surfaces that are not currently visible.  If a
+    /// frame callback is requested prior to committing a surface, the client should avoid drawing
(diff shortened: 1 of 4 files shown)


src/compositor.rs
```diff
@@ -42,15 +42,24 @@ pub trait CompositorHandler: Sized {
...
         &mut self,
         conn: &Connection,
         qh: &QueueHandle<Self>,
         surface: &wl_surface::WlSurface,
         time: u32,
-    );
+    ) {
+        // avoid unused warnings
+        let _ = (conn, qh, surface, time);
+    }
```
kchibisov, line 62: What's the reason to have default impl for this function?

danieldg: Frame callbacks aren't a mandatory part of dealing with surfaces, so forcing users to implement a possible no-op callback isn't useful.

danieldg: I actually think there's a bunch of trait functions that should have default no-op implementations, so splitting that out into its own PR and discussing it there may be better (unless it's unwanted for all).

kchibisov: I mean, the users could at least know that they may be need this function, since no one ever read the docs. And no, frame callback based drawing is pretty much mandatory if you want to have your client not glitch, especially with EGL.

wash2 (approved):

PR #421, opened 2023-10-19
ProvidesBoundGlobal API_VERSION fixes

danieldg:
Two recent bumps to maximum API versions did not also bump the implementations of ProvidesBoundGlobal even though API breaks were present.  This series bumps those API versions.  This is technically an API change, albeit a minor one that would only break code directly using the ProvidesBoundGlobal interface on one of these objects - and such code may already be broken in 0.18 because it will not handle the new events.

### DIFF
--- a/src/compositor.rs
+++ b/src/compositor.rs
@@ -69,6 +69,12 @@ pub struct CompositorState {
 }
 
 impl CompositorState {
+    /// The maximum API version for WlCompositor that this object will bind.
+    // Note: if bumping this version number, check if the changes to the wayland XML cause an API
+    // break in the rust interfaces.  If it does, be sure to remove other ProvidesBoundGlobal
+    // impls; if it does not, consider adding one for the previous (compatible) version.
+    pub const API_VERSION_MAX: u32 = 6;
+
     pub fn bind<State>(
         globals: &GlobalList,
         qh: &QueueHandle<State>,
@@ -76,7 +82,7 @@ impl CompositorState {
     where
         State: Dispatch<wl_compositor::WlCompositor, GlobalData, State> + 'static,
     {
-        let wl_compositor = globals.bind(qh, 1..=6, GlobalData)?;
+        let wl_compositor = globals.bind(qh, 1..=Self::API_VERSION_MAX, GlobalData)?;
         Ok(CompositorState { wl_compositor })
     }
 
@@ -185,7 +191,10 @@ pub struct Surface(wl_surface::WlSurface);
 
 impl Surface {
     pub fn new<D>(
-        compositor: &impl ProvidesBoundGlobal<wl_compositor::WlCompositor, 6>,
+        compositor: &impl ProvidesBoundGlobal<
+            wl_compositor::WlCompositor,
+            { CompositorState::API_VERSION_MAX },
+        >,
         qh: &QueueHandle<D>,
     ) -> Result<Self, GlobalError>
     where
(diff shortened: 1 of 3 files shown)


src/compositor.rs
```diff
@@ -185,7 +185,7 @@ pub struct Surface(wl_surface::WlSurface);
 
 impl Surface {
     pub fn new<D>(
-        compositor: &impl ProvidesBoundGlobal<wl_compositor::WlCompositor, 5>,
+        compositor: &impl ProvidesBoundGlobal<wl_compositor::WlCompositor, 6>,
```
kchibisov, line 188: Could you refactor the `6` into the version constant and use the constant in all places we use the version.

src/shell/xdg/mod.rs
```diff
@@ -168,7 +168,7 @@ pub struct XdgPositioner(xdg_positioner::XdgPositioner);
 
 impl XdgPositioner {
     pub fn new(
-        wm_base: &impl ProvidesBoundGlobal<xdg_wm_base::XdgWmBase, 4>,
+        wm_base: &impl ProvidesBoundGlobal<xdg_wm_base::XdgWmBase, 5>,
```
kchibisov, line 171: same about constants here.

kchibisov (changes requested):

src/shell/xdg/mod.rs
danieldg: I'm less sure about making the change to the uses that require an interface instead of implementing it. Making it a constant will cause the documentation to refer to that constant instead of the version number, and the entire reason this number exists is so that you have to bump it if there is an API break.

kchibisov: yes, but we have a situation where everyone forgot to bump anything and we have an API break.

danieldg: That's an argument for tying the bind and the trait impl together, but not an argument for also tying generic bounds to the constant.  The generic bounds being wrong would cause compile errors more directly.

kchibisov: I mean, bumping the protocol version is already a breaking change most of the time, it's not like you can miss it that easily.

The only way is to not have this bounds at all, so we never have a breaking change.

i509VCB:
If [variant_count](https://doc.rust-lang.org/std/mem/fn.variant_count.html) were stable, a compile time assertion could be done to see when a break occurs, but alas it is not...

danieldg:
> If [variant_count](https://doc.rust-lang.org/std/mem/fn.variant_count.html) were stable

Since smithay also controls the generation of these enums via a macro, we could add constants associated with the enum about the parts of API compatibility that can be auto-detected.  However, some breaks can't be detected that way (wl_compositor version 5 adding restrictions to `attach` is an example).  Not sure it would be all that easy to use said autodetection either - you'd need a compile-time assert inside the ProvidesBoundGlobal impl or something like that.

kchibisov:
>However, some breaks can't be detected that way (wl_compositor version 5 adding restrictions to attach is an example).

My reasoning for that is that the person doing the update to the protocol knows what changed in the protocol and can determine the behavior or the reviewer should. However getting on the hunt for constants across the codebase because we have _raw_ numbers which are not covered by any test case is what already happened.

It's not hard to identify a breakage, really. I don't even remember a single time thing like that happened and you can still `yank` due to breakage, however releasing a new breaking version is not fun at all, since it requires updating all the library crates extending SCTK functionality, where coordination will likely be solely on me.

kchibisov:
You can, alternatively, add doc tests if you still think that using `constants` instead of _raw_ numbers is good idea in such a case.

wash2 (approved):

wash2:
I guess I'll merge this and publish a new version for 0.18.1?

PR #426, opened 2023-10-31
Implement support for xdg-activation

Drakulix:
Implements xdg-activation.

The second commit adds it to the simple-window example as a test. Given how heavily commented that is, I guess that is not really a good place to put it. But without a proper window it doesn't make a whole lot of sense to use it. I could also just drop that commit?

### DIFF
--- a/src/activation.rs
+++ b/src/activation.rs
@@ -1,0 +1,200 @@
+use wayland_client::{
+    globals::{BindError, GlobalList},
+    protocol::{wl_seat, wl_surface},
+    Dispatch, Proxy, QueueHandle,
+};
+use wayland_protocols::xdg::activation::v1::client::{xdg_activation_token_v1, xdg_activation_v1};
+
+use crate::{
+    error::GlobalError,
+    globals::{GlobalData, ProvidesBoundGlobal},
+};
+
+/// Minimal implementation of [`RequestDataExt`].
+///
+/// Use a custom type implementing [`RequestDataExt`] to store more data with a token request
+/// e.g. to identify which request produced which token.
+#[derive(Debug, Clone)]
+pub struct RequestData {
+    /// App_id of the application requesting the token, if applicable
+    pub app_id: Option<String>,
+    /// Seat and serial of the window requesting the token, if applicable.
+    ///
+    /// *Warning*: Many compositors will issue invalid tokens for requests without
+    /// recent serials. There is no way to detect this from the client-side.
+    pub seat_and_serial: Option<(wl_seat::WlSeat, u32)>,
+    /// Surface of the window requesting the token, if applicable.
+    ///
+    /// *Warning*: Many compositors will issue invalid tokens for requests from
+    /// unfocused surfaces. There is no way to detect this from the client-side.
+    pub surface: Option<wl_surface::WlSurface>,
+}
+
+/// Data attached to a token request
+pub trait RequestDataExt: Send + Sync {
+    /// App_id of the application requesting the token, if applicable
+    fn app_id(&self) -> Option<&str>;
+    /// Seat and serial of the window requesting the token, if applicable.
+    ///
+    /// *Warning*: Many compositors will issue invalid tokens for requests without
+    /// recent serials. There is no way to detect this from the client-side.
+    fn seat_and_serial(&self) -> Option<(&wl_seat::WlSeat, u32)>;
+    /// Surface of the window requesting the token, if applicable.
+    ///
+    /// *Warning*: Many compositors will issue invalid tokens for requests from
+    /// unfocused surfaces. There is no way to detect this from the client-side.
+    fn surface(&self) -> Option<&wl_surface::WlSurface>;
+}
+
+impl RequestDataExt for RequestData {
+    fn app_id(&self) -> Option<&str> {
+        self.app_id.as_deref()
+    }
+
+    fn seat_and_serial(&self) -> Option<(&wl_seat::WlSeat, u32)> {
+        self.seat_and_serial.as_ref().map(|(seat, serial)| (seat, *serial))
+    }
+
+    fn surface(&self) -> Option<&wl_surface::WlSurface> {
+        self.surface.as_ref()
+    }
+}
+
+/// Handler for xdg-activation
+pub trait ActivationHandler: Sized {
+    /// Data type used for requesting activation tokens
+    type RequestData: RequestDataExt;
+    /// A token was issued for a previous request with `data`.
+    fn new_token(&mut self, token: String, data: &Self::RequestData);
+}
+
+/// State for xdg-activation
+#[derive(Debug)]
+pub struct ActivationState {
+    xdg_activation: xdg_activation_v1::XdgActivationV1,
+}
+
+impl ActivationState {
+    /// Bind the `xdg-activation` global
+    pub fn bind<State>(
+        globals: &GlobalList,
+        qh: &QueueHandle<State>,
+    ) -> Result<ActivationState, BindError>
+    where
+        State: Dispatch<xdg_activation_v1::XdgActivationV1, GlobalData, State> + 'static,
+    {
+        let xdg_activation = globals.bind(qh, 1..=1, GlobalData)?;
+        Ok(ActivationState { xdg_activation })
+    }
+
+    /// Activate a surface with the provided token.
+    pub fn activate<D>(&self, surface: &wl_surface::WlSurface, token: String) {
+        self.xdg_activation.activate(token, surface)
+    }
+
+    /// Request a token for surface activation.
+    ///
+    /// To attach custom data to the request implement [`RequestDataExt`] on a custom type
+    /// and use [`Self::request_token_with_data`] instead.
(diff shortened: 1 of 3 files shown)


src/activation.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// Seat and serial of the window requesting the token, if applicable.
+    ///
+    /// *Warning*: Many compositors will issue invalid tokens for requests without
+    /// recent serials. There is no way to detect this from the client-side.
+    fn serial(&self) -> Option<(&wl_seat::WlSeat, u32)>;
+    /// Surface of the window requesting the token, if applicable.
+    ///
+    /// *Warning*: Many compositors will issue invalid tokens for requests from
+    /// unfocused surfaces. There is no way to detect this from the client-side.
+    fn surface(&self) -> Option<&wl_surface::WlSurface>;
```
kchibisov, line 26: No need to use `Option` for both of them, just enforce the use of what should be passed..

src/activation.rs
```diff
@@ -0,0 +1,158 @@
...
+        self.0.as_ref().map(|(seat, serial)| (seat, *serial))
+    }
+
+    fn surface(&self) -> Option<&wl_surface::WlSurface> {
+        self.1.as_ref()
+    }
+
+    fn app_id(&self) -> Option<String> {
+        self.2.clone()
+    }
```
kchibisov, line 42: User can clone if they want to.

```suggestion
    fn app_id(&self) -> Option<&str> {
        self.2.as_ref()
    }
```

src/activation.rs
```diff
@@ -0,0 +1,155 @@
...
+
+/// Data attached to a token request
+pub trait RequestData: Send + Sync {
+    /// App_id of the application requesting the token, if applicable
+    fn app_id(&self) -> Option<String>;
+    /// Seat and serial of the window requesting the token, if applicable.
+    ///
+    /// *Warning*: Many compositors will issue invalid tokens for requests without
+    /// recent serials. There is no way to detect this from the client-side.
+    fn serial(&self) -> Option<(&wl_seat::WlSeat, u32)>;
```
kchibisov, line 21: You're returning `seat` and `serial` from a function called `serial`, sounds really odd.

I'd suggest to separate `seat` and `serial`. Also `&Wlseat` should be better.

src/activation.rs
```diff
@@ -0,0 +1,155 @@
...
+        self.2.clone()
+    }
+}
+
+/// Handler for xdg-activation
+pub trait ActivationHandler: Sized {
+    /// Data type used for requesting activation tokens
+    type RequestData: RequestData;
+    /// A token was issued for a previous request with `data`.
+    fn new_token(&mut self, data: &Self::RequestData, token: String);
```
kchibisov, line 50: token with data sounds like the order should

`fn new_token(&mut self, token: String, data)`.

kchibisov (changes requested):

src/activation.rs
Drakulix: Yes there is use in requesting a token without providing these.

src/activation.rs
Drakulix: That's the naming the protocol uses and it expects either both or none of the above. So I can't really make two methods out of it. We *could* just change the name.

src/activation.rs
kchibisov: I mean, what could it be, it's all non-documented. You better use `app_id` if you want to have a special logic in compositor.

src/activation.rs
Drakulix: sure, done.

src/activation.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Surface of the window requesting the token, if applicable.
+    ///
+    /// *Warning*: Many compositors will issue invalid tokens for requests from
+    /// unfocused surfaces. There is no way to detect this from the client-side.
+    fn surface(&self) -> Option<&wl_surface::WlSurface>;
+}
+
+impl RequestData
+    for (Option<(wl_seat::WlSeat, u32)>, Option<wl_surface::WlSurface>, Option<String>)
+{
```
kchibisov, line 31: Why do we even have this trait impl inside the `SCTK` sounds just wrong. Users should wrap the data if they want to, since they also will pass user data along the request. Winit does that for example to throttle requests.

kchibisov (changes requested):

src/activation.rs
Drakulix: E.g. requesting a token without an active surface. Which might be valid for privileged clients.

Because it is all non-documented, we *should* actually expose everything the protocol allows instead of assuming a certain compositor policy.

src/activation.rs
Drakulix: Yes, it is public api, but the only consumer is likely the xdg-activation implementation. Anyway changing this.

src/activation.rs
Drakulix: this better?

src/activation.rs
kchibisov: bump.

Drakulix: I can move this into the second commit. It felt to me like some simpler clients might not need any additional identifying data, when requesting a token, and would profit from this implementation.

kchibisov: As alternative, SCTK has this weird design where you ask without user data and with `with_user_data` functions and have a default user data like

```
struct XdgActiovTokenData {
   serial,
   seat,
   app_id,
   surface
}
```

And the trait like  `XdgActiovationTokenDataExt` so you can get it from the `U: XdgActivationTokenDataExt`.

I'm not quite sure what I like more, but the thing is that the current way is to do like I suggested. You can look at other APIs for that matter.

kchibisov: See https://docs.rs/smithay-client-toolkit/latest/smithay_client_toolkit/seat/keyboard/trait.KeyboardDataExt.html

src/activation.rs
Drakulix: Well, if you tell me what the implementation is supposed to do, if it gets either but not the other... just don't set it? That feels wrong, if we can enforce it on a type level.

src/activation.rs
Drakulix: Thanks for the hint, I wasn't aware of this. I'll look into that.

Drakulix: done

src/activation.rs
```diff
@@ -0,0 +1,180 @@
...
+    error::GlobalError,
+    globals::{GlobalData, ProvidesBoundGlobal},
+};
+
+#[derive(Debug, Clone)]
+pub struct RequestData {
+    pub app_id: Option<String>,
+    pub seat_and_serial: Option<(wl_seat::WlSeat, u32)>,
+    pub surface: Option<wl_surface::WlSurface>,
+}
```
kchibisov, line 18: I think it should point to docs on at least `RequestDataExt`, otherwise how the user would know what the fields should do?

src/activation.rs
```diff
@@ -0,0 +1,180 @@
...
+    ///
+    /// *Warning*: Many compositors will issue invalid tokens for requests without
+    /// recent serials. There is no way to detect this from the client-side.
+    fn seat_and_serial(&self) -> Option<(&wl_seat::WlSeat, u32)>;
+    /// Surface of the window requesting the token, if applicable.
+    ///
+    /// *Warning*: Many compositors will issue invalid tokens for requests from
+    /// unfocused surfaces. There is no way to detect this from the client-side.
+    fn surface(&self) -> Option<&wl_surface::WlSurface>;
+}
```
kchibisov, line 34: The whole concept of that `with_data` methods is that you ask for `RequestData` from the `Ext` not for the fields, the fields are stored on the data you use inside the `SCTK`.

So all your docs should go to the `RequestData`, but this trait should just have `fn request_data(&self) -> &RequestData`. As you can see from the `KeyboardDataExt`.

Then you don't need to have associated type in user code as well iirc.

src/activation.rs
Drakulix: Docs updated

src/activation.rs
Drakulix: > Then you don't need to have associated type in user code as well iirc.

I think I would still need to that to know what type the `RequestDataExt` is to return it in the `ActivationHandler`. The only thing that returning `RequestData` from the trait instead gives is easier composability, as the user may just store the `RequestData` in it's type. But they can already do that and forwarding three methods doesn't seem much less ergonomic to me.

Anyway, sure I can change that, but I don't see how this gets rid of the associated type.

kchibisov: Ah, yes, it won't get it because you pass the `data`, SCTK usually passes the wayland object and user gets the data. It's a bit meh, but it's consistent.

Drakulix: > SCTK usually passes the wayland object and user gets the data. It's a bit meh, but it's consistent.

Uhh, that feels entirely unnecessary given the user has never seen a token object up until this point.

kchibisov: yeah, you can leave mostly as is, it's just sctk has not that great api with delegate and user data, it's very verbose.

wash2 (approved):

PR #437, opened 2024-01-18
Update `wgpu` example to `wgpu` 0.19

ids1024:


ids1024:
Hm, so wgpu requires Rustc 1.70?

It's not really a problem for the example to require a higher MSRV than the library, but it's a bit awkward to CI test that.

i509VCB:
> Hm, so wgpu requires Rustc 1.70?
> 
> It's not really a problem for the example to require a higher MSRV than the library, but it's a bit awkward to CI test that.

What wgpu uses is largely based on Firefox.

I'd be fine bumping to 1.70. Although didn't rwh bump up and I know ash now needs 1.69 on git master.

examples/wgpu.rs
```diff
@@ -231,6 +215,7 @@ impl WindowHandler for Wgpu {
             alpha_mode: wgpu::CompositeAlphaMode::Auto,
             width: self.width,
             height: self.height,
+            desired_maximum_frame_latency: 2,
```
i509VCB, line 218: Where is this field being used?

i509VCB (approved):

examples/wgpu.rs
ids1024: This is a field of `wgpu::SurfaceConfiguration`. That used to be hard-coded to 2, but is now configurable.

i509VCB: Oh okay so the web UI probably didn't show the struct owning the field.

PR #443, opened 2024-02-25
Require an explicit unlock call in SessionLock

danieldg:
The entire reason unlock_and_destroy is distinct from destroy is to avoid accidental unlock operations.  The current SCTK implementation carefully undoes this work by unlocking on drop, even if that drop is due to a panic or careless drop of the lock object.

Add an explicit unlock call so that screen lockers can indicate when they intend to unlock (generally after some kind of authentication has happened).


wash2 (approved):
I can see why it would be a problem if unlock is called automatically when the application panics.

ids1024:
The ideal solution would be to have some kind of "linear typing" which would prevent implicit dropping except though panics. But yeah, this is probably correct without any better solution.

This is a breaking API change for any client using session lock.

PR #444, opened 2024-02-28
Work around touch up events delivered too late with certain Weston versions

tronical:
Weston <= 10 at least has a bug where the frame event isn't always sent when the last touch point was released.

This is tracked in
https://gitlab.freedesktop.org/wayland/weston/-/issues/44 and it would seem that
https://gitlab.freedesktop.org/wayland/weston/-/commit/5448580111b5ff992ce2603cb6e99b9f54db7ad8 may fix it.

Meanwhile, work around the issue by processing all buffered events when there are no more active touch points.

### DIFF
--- a/src/seat/touch.rs
+++ b/src/seat/touch.rs
@@ -30,6 +30,7 @@ impl TouchData {
 #[derive(Debug, Default)]
 pub(crate) struct TouchDataInner {
     events: Vec<TouchEvent>,
+    active_touch_points: Vec<i32>,
 }
 
 #[macro_export]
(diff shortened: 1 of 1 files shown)


tronical:
Related downstream issue on our side that triggered this: https://github.com/slint-ui/slint/issues/4551

kchibisov:
No, adding workarounds for protocol violation is not how it should be done, I'd suggest to fix weston instead of someone relies on it. The protocol mandates the processing on the frame event, no frame event means that the event was likely canceled (though cancel could be send as well), which could introduce bugs along the way.

tronical:
The problem is that upgrading Weston is not always an option on embedded systems :( - especially when they're based on older Yocto or Debian versions.

kchibisov:
>I also wouldn't mind applying this workaround on a different level, but I don't know if that's possible.

I'm not sure it's because it breaks core protocol. Though, the only thing that could be broken is the `cancel` event, but cancel event implies that all the touch points are lifted, thus probably it won't really break anything when this workaround will be applied.

So maybe it's fine, but I'll leave it up to @wash2 .

wash2:
I think this is unlikely to break anything, so it's an ok workaround.

kchibisov:
Well, yeah, the only area where it can break, is `wl_touch::cancel`, but it's likely won't break because it's the last point and you need to lift everything anyway.

wash2 (approved):

PR #448, opened 2024-04-11
add way to make subsurfaces from surfaces

maxhbooth:
I'd like to be able to pass an existing surface and make it a subsurface, instead of using the surface created in `create_subsurface`.  Would that be possible with something like this?   Thanks :)

### DIFF
--- a/src/subcompositor.rs
+++ b/src/subcompositor.rs
@@ -42,6 +42,21 @@ impl SubcompositorState {
             self.subcompositor.get_subsurface(&surface, &parent, queue_handle, subsurface_data);
         (subsurface, surface)
     }
+
+    pub fn subsurface_from_surface<State>(
+        &self,
+        surface: &WlSurface,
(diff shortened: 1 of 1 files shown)


wash2:
What exactly is the use case for this? If the passed surface already has a role, this will be a protocol error, and a surface cannot be its own parent anyway, or else you get a protocol error. https://wayland.app/protocols/wayland#wl_subcompositor:request:get_subsurface

wash2:
> I know that my surfaces won't have a role yet. I'd be happy to add a check for that, but I don't think we have enough information here to know.

Ok, gotcha.

 I think the function might need a better name because currently it's a bit hard to tell without checking the implementation if it is creating a subsurface that is a child of the passed surface, or if it is creating a subsurface from the passed surface. Maybe `subsurface_from_surface`?

It could possibly return an `Option` so that it doesn't have to unwrap on the parent surface.

wash2 (approved):

PR #450, opened 2024-05-03
expose xdg_decoration_manager in xdg_shell

maxhbooth:
I'd like to be able to test if this interface is present.


kchibisov:
configure will tell you that you have `ClientOnly` when it's not present.

PR #456, opened 2024-06-06
Add wl shell protocol implementation

lmaxyz:
Make available using wl shell protocol for windows creation.

## Important

This implementation is only for make available to create windows in Sailfish OS by `winit`.
I don't know what else it is need for, because it's deprecated and xdg uses everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Add wl shell protocol implementation
+
 ## 0.19.1 - 2024-06-04
 
 #### Additions
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -32,13 +32,13 @@
 //!
 //! The layer shell may be used to implement many desktop shell components, such as backgrounds, docks and
 //! launchers.
-//! 
+//!
 //! ## Wl shell
-//! 
+//!
 //! The Wl shell is a deprecated protocol which is like xdg shell but less functional.
-//! 
+//!
 //! ### Why use the Wl shell
-//! 
+//!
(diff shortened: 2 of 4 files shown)


i509VCB:
I'm not terribly interested in wl_shell support. But I guess someone needs it.

For winit you'll need to convince @kchibisov, just making a pull request here doesn't guarantee winit will implement it.

kchibisov:
We explicitly dropped `wl_shell` because no one should really use it, it's dropped in mostly every compositor. I personally don't see a reason in using it at all at this point and rather like things to move to `xdg_shell` instead.

lmaxyz:
I see. But compositor in Sailfish OS still uses `wl_shell`, so i decided to add its support to make available creating application using slint ui for this OS.
This OS also uses `qt_extended_surface` protocol to correctly close its applications and I'm also going to add implementation of this protocol to `winit`, despite it's deprecated too.

If it's a big problem to have this code in repository, I'll fork it. But I think it's not a big deal to implement it as feature, that will uses only if needs, and xdg will use by default.

kchibisov:
I'm not interested in maintaining protocols no one can really test in winit. I'd suggest to update sailfish instead.

lmaxyz:
And why do you think nobody can test these protocols?

I agree with your suggestion, but I only do what I can do, and `Sailfish OS` with its c++ code out of my opportunities, so I decided to add support for these protocols to `winit` until sailfish's maintainers updates compositor.

So, you won't accept PR for these protocols for `winit`, right?

kchibisov:
I won't accept them for winit, since all mainline compositors don't support them anymore, meaning that average contributor can not test them, because they require special setups, etc.

PR #457, opened 2024-06-19
Update `calloop` to `0.14.0`, `calloop-wayland-source`

ids1024:
Depends on https://github.com/Smithay/calloop-wayland-source/pull/4.

I guess this will be a breaking change... hopefully `calloop` can be more stable now.


notgull:
> I guess this will be a breaking change... hopefully `calloop` can be more stable now.

Barring any unforeseen developments I plan to make `calloop` v0.14.0 into version v1.0.0. I even plan on using the [semver trick](https://github.com/dtolnay/semver-trick) to make the upgrade as smooth as possible.

wash2 (approved):

PR #467, opened 2024-09-04
Add Clone for dmabuf feedback structs

chrisduerr:
The `DmabufHandler::dmabuf_feedback` handler provides access to DMA buffer feedback, however working with this feedback becomes difficult when it needs to be stored in multiple places due to missing derives.

So this patch adds the automatic `Clone` derive so the existing types in SCTK can be reused downstream.

The `DmabufFeedback` itself does not implement `Clones` since it relies on the `Mmap` type, which is not cloneable.

---

Note I didn't actually require this myself since I was able to clone the WebKit type I used it for. But it seems useful for other people so I thought I might as well put up a PR.

### DIFF
--- a/src/dmabuf.rs
+++ b/src/dmabuf.rs
@@ -19,7 +19,7 @@ type dev_t = u64;
 use libc::dev_t;
 
 /// A preference tranche of dmabuf formats
-#[derive(Debug)]
+#[derive(Clone, Debug)]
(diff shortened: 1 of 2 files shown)


PolyMeilex (approved):

PolyMeilex:
Sounds like a good idea, sorry for the delay.

PR #470, opened 2024-10-01
WIP image-capture-source / image-copy-capture

ids1024:
There may be a benefit to a higher-level API, perhaps something async based, but that probably can go in a different crate.

I still need to fix a few things here and get a working test client.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ thiserror = "1.0.30"
 wayland-backend = "0.3.0"
 wayland-client = "0.31.1"
 wayland-cursor = "0.31.0"
-wayland-protocols = { version = "0.32.1", features = ["client", "staging", "unstable"] }
+wayland-protocols = { version = "0.32.4", features = ["client", "staging", "unstable"] }
 wayland-protocols-wlr = { version = "0.3.1", features = ["client"] }
 wayland-scanner = "0.31.0"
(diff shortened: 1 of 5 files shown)


HigherOrderLogic:
Hi, I'm interested in this PR. From what I can see, seems like this one is abandoned. Mind if I open a new one based on your changes?

ids1024:
The current version of this is in https://github.com/pop-os/cosmic-protocols/tree/main/client-toolkit/src/screencopy, with the addition of a COSMIC-specific protocol for Workspace capture.

It would be good to get this in `smithay-client-toolkit` though. On the client side it should be easy enough to accommodate extension protocols for new capture sources outside of sctk.

HigherOrderLogic:
So should I just copy straight from Cctk (minus the Cosmic protocol) or is there any changes you wanna make?

ids1024:
Originally the cosmic version used a private protocol (before the image-copy-capture protocol was finalized in `wayland-protocols`, then for a while it supported both until everything was updated.

The way https://github.com/pop-os/cosmic-protocols/blob/main/client-toolkit/src/screencopy/capture_source.rs defines `CaptureSourceKind` and `CaptureSource`  doesn't quite work if we want it to be extensible, so the API for capture sources needs to be a bit different. But shouldn't be that complicated.

PR #486, opened 2025-02-14
Fix ci

wash2:


.cirrus.yml
```diff
@@ -1,10 +1,10 @@
 task:
   only_if: $CIRRUS_BRANCH == 'master' || $CIRRUS_PR != ''
   matrix:
-    - name: FreeBSD 13.2
+    - name: FreeBSD 14.2
       freebsd_instance:
-        image_family: freebsd-13-2-snap
- # Install Rust
+        image_family: freebsd-14-2-snap
```
ids1024, line 6: `gcloud compute images list --no-standard-images --project=freebsd-org-cloud-dev` shows there's a `freebsd-13-4-snap` image (the oldest version listed), so it can use that if we want to test older than FreeBSD 14. That's what I've updated `wayland-rs` to.

I guess that's reasonable to treat as the minimum supported FreeBSD version (https://en.wikipedia.org/wiki/FreeBSD_version_history#Version_history lists it as the oldest "maintained" version, assuming that's accurate). I guess I can update https://github.com/Smithay/client-toolkit/pull/341 if FreeBSD 12 support doesn't matter now.

PolyMeilex (approved):

PR #489, opened 2025-03-29
Support tablet-v2

### LINKED ISSUE #139: Support for graphics tablets
It seems like graphics tablets are currently unsupported. There is a protocol extension https://docs.rs/wayland-protocols/0.27.0/wayland_protocols/unstable/tablet/index.html

chris-morgan:
This is still a work in progress. Some remarks:

- I’ve got a different pattern going on from the normal udata, and I think it’s noticeably nicer in some regards, but it’s also probably a bad idea. Some remarks are present in commit messages and maybe comments.
- I’ve named things somewhat differently from upstream, broadly speaking `foo::{X, Y, Z}` because `foo::{FooX, FooY, FooZ}` gets tedious here. *I* think it’s an improvement, see how it looks in the example. Honestly I’d suggest doing this more generally, but I think it’s justifiable even as a tablet-local convention, because the names get unwieldy.
- There are still various TODOs, some of which I would appreciate review on.
- Tablet pads are not supported (but are plumbed just enough that their presence *shouldn’t* crash it).
- Themed cursors need to be worked in.

Took me a while to figure out various things, and on account of working various things out I went back and changed lots more, but it’s reached the stage where I’m mostly happy with it. The commit messages will at times be useful, but the history is inevitably a bit of a mess, for such an exploratory thing (this was my first tangle with programming against Wayland).

Things work reliably for me on my XP-Pen Artist 16 (2nd Gen).

Fixes #139.

Now I can get on with making my own app that uses this…


src/seat/tablet.rs
```diff
@@ -0,0 +1,90 @@
...
+        conn: &Connection,
+        qh: &QueueHandle<Self>,
+        tablet: &ZwpTabletV2,
+        info: Info,
+    );
+
+    /// Sent when the tablet has been removed from the system.
+    /// When a tablet is removed, some tools may be removed.
+    ///
+    /// This method is responsible for running `tablet.destroy()`.  ← TODO: true or not?
```
wash2, line 25: It would be a good time to destroy the object, but I suppose it isn't strictly responsible for it.

wash2 (commented):
- I’ve got a different pattern going on from the normal udata, and I think it’s noticeably nicer in some regards, but it’s also probably a bad idea. Some remarks are present in commit messages and maybe comments.

If I've understood correctly, I guess the main drawback I think of is that the info is only available in the event really, instead of via a method on the `Data`? So, users would have to add this to their state if they want to use it after the info method. I don't think it is too bad.

> I’ve named things somewhat differently from upstream, broadly speaking foo::{X, Y, Z} because foo::{FooX, FooY, FooZ} gets tedious here. I think it’s an improvement, see how it looks in the example. Honestly I’d suggest doing this more generally, but I think it’s justifiable even as a tablet-local convention, because the names get unwieldy.

Ya, for now it would be fine as just a tablet-local convention.

> Tablet pads are not supported (but are plumbed just enough that their presence shouldn’t crash it).

This looks good to me, but unfortunately I don't have anything to test it with, or really the rest of the PR, to be honest.

PR #495, opened 2025-05-23
Implement experimental input method protocol

dcz-self:
With popup additions: https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/407

Missing/untested: repositioning.

### DIFF
--- a/src/seat/input_method_v3.rs
+++ b/src/seat/input_method_v3.rs
@@ -1,0 +1,744 @@
+/*! This implements support for the experimental xx-input-method-v2 protocol.
+ * That protocol will hopefully become -v3 without changing the API at some point.
+ */
+
+use crate::compositor::Surface;
+use crate::globals::GlobalData;
+
+use log::{debug, warn};
+
+use std::collections::HashMap;
+use std::num::Wrapping;
+use std::ops::Deref;
+use std::sync::{Arc, Mutex, MutexGuard, Weak};
+
+use wayland_client::globals::{BindError, GlobalList};
+use wayland_client::protocol::wl_seat::WlSeat;
+use wayland_client::protocol::wl_surface;
+use wayland_client::WEnum;
+use wayland_client::{Connection, Dispatch, Proxy, QueueHandle};
+use wayland_protocols::wp::text_input::zv3::client::zwp_text_input_v3::{
+    ChangeCause, ContentHint, ContentPurpose,
+};
+
+use wayland_protocols_experimental::input_method::v1::client as protocol;
+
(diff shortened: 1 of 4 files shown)


dcz-self:
To test:
- anvil from https://github.com/Smithay/smithay/pull/1745
- stiwri from "popup" branch https://codeberg.org/dcz/stiwri/src/branch/popup

```
# in anvil
cargo run --release -- --winit &
WAYLAND_DISPLAY=wayland-1 yad &
# in stiwri
WAYLAND_DISPLAY=wayland-1 cargo run --release --bin im_popup
```

dcz-self:
Repositioning has been tested now:

```
WAYLAND_DISPLAY=wayland-1 cargo run --release --bin im_popup_reposition
```
[Screencast_20250524_202116.webm](https://github.com/user-attachments/assets/471d5759-2344-4b51-99a7-2814bc8c2d2f)

src/seat/input_method_v3.rs
```diff
@@ -0,0 +1,711 @@
...
+    input_method: XxInputMethodV1,
+}
+
+impl InputMethod {
+    pub fn input_method(&self) -> &XxInputMethodV1 {
+        &self.input_method
+    }
+
+    pub fn set_preedit_string(&self, text: String, cursor: CursorPosition) {
+        // TODO: should this enforce indices on codepoint boundaries?
```
wash2, line 154: I think that would be good. It could be changed to return a Result, with an error when there is an index not on a boundary?

src/seat/input_method_v3.rs
```diff
@@ -0,0 +1,711 @@
...
+        };
+        self.input_method.set_preedit_string(text, start, end)
+    }
+
+    pub fn commit_string(&self, text: String) {
+        self.input_method.commit_string(text)
+    }
+
+    pub fn delete_surrounding_text(&self, before_length: u32, after_length: u32) {
+        // TODO: this has 2 separate behaviours:
```
wash2, line 173: I'm not sure that I understand the difference in the protocol. The preedit text is just removed in the case that it is present, and otherwise it works the same, right?

src/seat/input_method_v3.rs
dcz-self: done.

dcz-self:
Bump. Anyone interested?

wash2:
Hi, sorry for the delayed response. Are you planning on depending on https://github.com/Smithay/wayland-rs for the protocol, instead of your repository? I think it otherwise looks good.

PR #497, opened 2025-07-09
Add wl_pointer extensions from v8 and v9

dcz-self:
I've done a little testing on it.
The user of the API is responsible for choosing which event they want to rely on.


wash2 (approved):

dcz-self:
That was easier than expected, thank you!

PR #498, opened 2025-07-09
keyboard: Fix sending press events when repeat is disabled

dcz-self:
When repeat was disabled, the press events were being eaten.

I think also every situation where the timer happens only once is affected.


dcz-self:
This now passes CI.

wash2 (approved):

PR #499, opened 2025-07-09
Fix CI

dcz-self:
The CI is set up in such a way that changes in external projects can fail runs that succeeded previously. This adds an exception for clippy to keep existing usages of the offending feature.


dcz-self:
I also added a fix for outdated repositories but I don't really understand where that one's coming from, so it might be overkill and at the same time insufficient.

PR #505, opened 2025-08-19
Allow extracting the Surface used to construct a Window.

nicolasavru:
Wayland allows reusing a surface for the same role it had previously.


nicolasavru:
I don't love the unwraps, so if you think of a better way to do this, please let me know.

PR #519, opened 2026-04-03
Prototype a `Dispatch2` trait to dispatch delegate macros to oblivion

ids1024:
I've though of some alternate ways to define [`Dispatch`](https://docs.rs/wayland-client/latest/wayland_client/trait.Dispatch.html) (as an alternative to the change in https://github.com/Smithay/smithay/pull/1327). Today I had an idea for a way to prototype such a thing outside of `wayland-rs`. Due to the orphan rule, we can't create any blanket impls of `Dispatch` that are generic over the `State` the trait is implemented for (unless our crate is the one that defined the type `I`). But we can provide a `delegate_dispatch2!` macro that will add such a blanket impl for a particular concrete type.

`Dispatch2` is similar to `wayland_client::Dispatch`, but it's now implemented for the object user data type instead of the state type, uses `&self` for the argument pointer to the user data, and removes the extra `State = Self` type bound.

Implementing this for the user data seems reasonable, since we already need to pass a user data when sending a request that creates an object. So the type passed there indicates what dispatch implementation to use. (It would also work to use `State: Dispatch<UserData, I>` instead of `UserData: Dispatch<I, State>`; but it would *not* work to use `State: Dispatch<I, UserData>` where `I` is a type from a different crate and `State` is generic. That isn't allowed under the orphan rule. So of the alternatives that work, this draft uses the one that seemed best to me.)

This draft uses `Dispatch2` and eliminates the dispatch macros used in the `simple_window` example, except:
- We can't make this generic over `KeyboardDataExt`, `PointerDataExt`, `SurfaceDataExt`, `RequestDataExt` (at least if this trait were part of wayland-rs, and not the same crate)
  * We can change the API to instead wrap the user data in a type provided by sctk. This would actually be potentially simpler anyway, since a trait impl wouldn't be needed. That isn't done here.
- For registries, we likewise can't implement for the `GlobalList` type, since that is a `wayland-client` type, not an sctk one

`event_created_child()` is unchanged, but we should think of better ways to handle that too.

It is still necessary to have `Dispatch` bounds in methods of the `*State` types that create objects, but not in the implementations of `Dispatch2`. Though type bounds shouldn't be needed anymore if this were moved to wayland-rs and replaced `Dispatch`.

If an API change like this is good, it could be incorporated in the next breaking update to wayland-rs. Does anyone see non-obvious issues with this?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        rust: ['1.71.0', 'stable', 'beta']
+        rust: ['1.71.1', 'stable', 'beta']
     runs-on: ubuntu-latest
 
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## Unreleased
+
+#### Breaking Changes
+- `*DataExt` traits removed; now `*Data` types wrap custom data
+  * Any calls to `Proxy::data` will need to be updated for new type
+- All `delegate_*!` macros except `delegate_registry!` replaced with single
+  `delegate_dispatch2!` in preparation for future wayland-rs release.
+
 ## 0.20.0 - 2025-07-29
 
 #### Breaking Changes
--- a/examples/data_device.rs
+++ b/examples/data_device.rs
@@ -11,16 +11,14 @@ use smithay_client_toolkit::reexports::calloop::{
 };
 use smithay_client_toolkit::reexports::calloop_wayland_source::WaylandSource;
 use smithay_client_toolkit::{
-    compositor::{CompositorHandler, CompositorState},
+    compositor::{CompositorHandler, CompositorState, FrameCallbackData},
     data_device_manager::{
         data_device::{DataDevice, DataDeviceHandler},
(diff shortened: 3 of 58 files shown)


ids1024:
Rebased on top of https://github.com/Smithay/client-toolkit/pull/524, so the `*DataExt` issue is addressed.

Not sure about how we want to handle `GlobalList`/`delegate_registry!`. It would okay to have a macro for just that, but a bit annoying.

ids1024:
I've cleaned up the history, added detailed commit messages, and added changelog entries.

Along with the examples and tests, I think https://github.com/rust-windowing/winit/pull/4567 verifies nothing seems to be wrong with how this works. And https://github.com/Smithay/client-toolkit/pull/525 confirms this seems to work smoothly with the wayland-rs release instead of providing a `Dispatch2` macro here.

I think that confirms I can finish and merge https://github.com/Smithay/wayland-rs/pull/902 and start moving towards actually releasing wayland-rs with these things.

`delegate_registry!` is kept here since I think that will be replaced using some change in wayland-rs (maybe a version of `ProvidesRegistryState` can be provided by `wayland-client`).

This could be left for merge until after wayland-rs is released if preferred, but I think this should be okay to merge now. I guess sctk does breaking API changes a bit less than smithay, but it should be probably sctk won't need to do a release before the wayland-rs update anyway.

wash2 (approved):

PR #521, opened 2026-04-20
re-export feature `client_system` from `wayland_backend`

hack3rmann:
This feature is needed by any project that would use Smithay client-toolkit to integrate with anything via FFI, like wgpu for instance.


ids1024:
`wayland-client` now offers a `system` feature that requires `wayland-backend/system`, so sctk can also just drop the explicit dependency on `wayland-backend` and use `wayland-client/system` here. The feature call also just be called `system` (the fact it is for clients would be implied.)

A feature flag isn't strictly necessary, since a consumer can add an explicit wayland-client or wayland-backend dependency, but seems appropriate given the crate is already re-exported in https://smithay.github.io/client-toolkit/smithay_client_toolkit/reexports/index.html.

hack3rmann:
1. Removed the explicit dependency on `wayland-backend`
2. Renamed the feature to `system`

Appreciate your concerns on the flag, but I tried using `wayland-backend` as an explicit dependency with `client_system` being enabled and it didn't work out for some reason -- maybe I did it the wrong way.
Also, it's more clean to have the flag IMHO. It tells that the crate provides raw FFIs indeed

wash2 (approved):

hack3rmann:
I saw there was 3 failed CI workflows:

1. Formatting issue (my bad, just solved)
2. ci (1.71.0) -- calloop v0.14.4 does need rustc 1.71.1 or newer. The pipeline fetches rust toolchain v1.71.0
3. Cirrus CI:
```
Not enough compute credits to prioritize tasks!
Failed to start an instance: INVALID_ARGUMENT: Snap images are not supported due to boot stability.
```

There's another pending pull request that requires issues `2` and `3` to be solved. Issue `3` could only be solved by the maintainer though issue `2` could be solved via pull-request.

Do I need to open issues for `2` and `3`?
What would be better: updating CI workflow's rust toolchain's version to 1.71.1 at least, or downgrading calloop to the nearest version that supports building with rustc 1.71.0?

wash2:
Updating the CI toolchain version would be good. I'll merge this after that and the last issue can be resolved in another PR.

hack3rmann:
MSRV updated

hack3rmann:
Opened another pull request for disabling Cirrus CI, because it's going to be shut down:
https://github.com/Smithay/client-toolkit/pull/522

PR #524, opened 2026-05-11
Udata `*Data` types wrap user provided `U` instead of having a `*DataExt` trait

ids1024:
For https://github.com/Smithay/client-toolkit/pull/519, we need to remove the `*DataExt` traits, and instead make `*Data` wrap `U` so we can provide blanket implementations of the future `wayland_client::Dispatch` 

This probably doesn't need to be merged before https://github.com/Smithay/client-toolkit/pull/519, which can then be based on this, but it seems good to keep these changes in clean separate commits, and verify everything works here. May as well have a draft PR before this part is done.

The various `get_keyboard_*` methods are made a bit more redundant in implementation here, but maybe they could now use a builder pattern.

This is a breaking change to code that uses `.data::<...>` to get the udata, since the type has changed. Though maybe it would be best to replace use of that with helper `from_resource()` methods as smithay has for a couple things.

### DIFF
--- a/examples/simple_window.rs
+++ b/examples/simple_window.rs
@@ -85,6 +85,7 @@ fn main() {
                 seat_and_serial: None,
                 surface: Some(window.wl_surface().clone()),
                 app_id: Some(String::from("io.github.smithay.client-toolkit.SimpleWindow")),
+                udata: (),
             },
         )
     }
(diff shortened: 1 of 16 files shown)


ids1024:
Changing the `*Data` types this way seems good overall, though it's awkward that the breaking change isn't caught by the compiler since `.data::<_>` will fail at runtime.

ids1024:
Closing since this has been incorporated into https://github.com/Smithay/client-toolkit/pull/519.

PR #529, opened 2026-06-28
Add ext_workspace_v1 helper

pastthepixels:
This code might be a bit jank since I'm not familiar with the codebase nor Wayland protocol and mostly modified code from the foreign_toplevel_list helper. Feedback would be appreciated!

This implements a helper for the ext_workspace_v1 protocol aka everything I could read up [here](https://wayland.app/protocols/ext-workspace-v1). This is good for e.g. graphical shells, like mine which has an implementation [here](https://codeberg.org/pastthepixels/gaypanel/src/commit/07c984b1452aee15359b5b9c40b63ee775c0c855/src/wayland/workspaces.rs)


src/workspace_manager.rs
```diff
@@ -0,0 +1,324 @@
...
+    ext_workspace_handle_v1::{self, State, WorkspaceCapabilities},
+    ext_workspace_manager_v1,
+};
+
+/// Information about a workspace.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+#[non_exhaustive]
+pub struct WorkspaceInfo {
+    // ID
+    pub id: String,
```
Ph4ntomas, line 17: IMO this should be an option, as the compositor may not send this event

src/workspace_manager.rs
```diff
@@ -0,0 +1,324 @@
...
+    // Name
+    pub name: String,
+    // State
+    pub state: Option<WEnum<State>>,
+    // Coordinates
+    pub coordinates: Vec<u8>,
```
Ph4ntomas, line 23: Coordinates are `u32`. The protocol sends it as an array of bytes, but it should be retransformed.

src/workspace_manager.rs
```diff
@@ -0,0 +1,324 @@
...
+    // Capabilities
+    pub capabilities: Option<WEnum<WorkspaceCapabilities>>,
```
Ph4ntomas, line 25: Both `state` and `capabilities` are bitfields, not enums, so IMO wrapping them in `WEnum` is wrong here.

Instead, it would be better to convert the enum back to the raw bitfield and store that (If more than one flag is set, the value will be stored as `WEnum::Unknown(<value>)`)

Since they are bitfield, wrapping them in `Option` is kinda superfluous too, as their default value (all flags unset) is a perfectly valid state.

src/workspace_manager.rs
```diff
@@ -0,0 +1,324 @@
...
+/// Information about a workspace group.
+#[derive(Clone, Debug, Default)]
+#[non_exhaustive]
+pub struct GroupInfo {
+    // Outputs
+    pub outputs: Vec<wl_output::WlOutput>,
+    // Workspaces
+    pub workspaces: Vec<ext_workspace_handle_v1::ExtWorkspaceHandleV1>,
+    // Capabilities
+    pub capabilities: Option<WEnum<GroupCapabilities>>,
```
Ph4ntomas, line 37: Same remark as `WorkspaceCapabilities`, this is also a bitfield, so `Option` and `WEnum` doesn't quite match here IMO.

src/workspace_manager.rs
```diff
@@ -0,0 +1,324 @@
...
+struct WorkspaceInner {
+    current_info: Option<WorkspaceInfo>,
+    pending_info: WorkspaceInfo,
+}
+
+#[derive(Debug, Default)]
+struct GroupInner {
+    current_info: Option<GroupInfo>,
+    pending_info: GroupInfo,
+}
```
Ph4ntomas, line 50: IMO `pending_info` should be wrapped in `Option` here.

There are two advantages to that:
- on `done`, you'd just need to do `self.current_info = pending_info.take()`
- the `done` handler could 'know' which workspace/group changed, and call the `WorkspaceHandler::update_*` method only for items whose state has actually changed.

src/workspace_manager.rs
```diff
@@ -0,0 +1,324 @@
...
+        workspace.data::<WorkspaceData>()?.0.lock().unwrap().current_info.clone()
+    }
+
+    /// Returns information about a workspace group.
+    pub fn info_group(
+        &self,
+        group: &ext_workspace_group_handle_v1::ExtWorkspaceGroupHandleV1,
+    ) -> Option<GroupInfo> {
+        group.data::<GroupData>()?.0.lock().unwrap().current_info.clone()
+    }
```
Ph4ntomas, line 103: While those function works, it would be nice to have a proper handle object for both workspace and groups, on which you could query the state directly.

On that note, the actual handles client code would get could be created after the initial `Done` event, meaning the committed state would not need to be behind an option, hiding that implementation detail.

src/workspace_manager.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    pub fn stop(&self) {
+        if let Ok(workspace_manager) = self.workspace_manager.get() {
+            workspace_manager.stop();
+        }
+    }
+}
```
Ph4ntomas, line 110: The `WorkspaceManager` should either expose a `pub fn commit(&self)` or the underlying `ExtWorkspaceManagerV1`

src/workspace_manager.rs
```diff
@@ -0,0 +1,324 @@
...
+    ) {
+        match event {
+            ext_workspace_manager_v1::Event::WorkspaceGroup { workspace_group } => {
+                state.ext_workspace_state().groups.push(workspace_group);
+            }
+            ext_workspace_manager_v1::Event::Workspace { workspace } => {
+                state.ext_workspace_state().workspaces.push(workspace);
+            }
+            ext_workspace_manager_v1::Event::Done => {
+                // TODO: is cloning really the best for performance?
```
Ph4ntomas, line 193: I think here, it would be best not to clone the whole workspace/group `Vec` and instead build a Vec containing only the changed value (see previous comment about having both pending state and current state being wrapped in option).

That way, you can still have the new/update functions, but you can call them at a point you no-longer need to hold the `&mut WorkspaceManager`.
For the initial burst of events, it makes no differences, but later on, you'd clone less items overall.

PR #534, opened 2026-07-14
v0.21.0

wash2:
After this is merged, I can make a tag and publish the update.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,22 @@
 
 ## Unreleased
 
+## 0.21.0 - 2026-07-14
+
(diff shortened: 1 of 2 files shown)


ids1024 (approved):
Hopefully it won't be too much longer now before the next wayland-rs release and an update to sctk with https://github.com/Smithay/client-toolkit/pull/525, which will need other breaking sctk release.

But I don't think that's really an issue. Other than temporarily adding a call to `delegate_dispatch2`, the changes needed for this release are an incremental step to what the next release will need.

PolyMeilex (approved):

